PR #1, opened 2019-10-10
Remove plugins whose manifests don't exist

sonicaj:
This PR removes plugins from INDEX who's manifests don't exist yet.


sonicaj:
@miwi-fbsd will be bringing in the manifests. Closing this

PR #2, opened 2019-10-10
- Added channels-dvr and homebrigdge mainfests

miwi-fbsd:


channels-dvr.json
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "channels-dvr",
+    "release": "11.2-RELEASE",
```
sonicaj: 11.3 ?

homebridge.json
```diff
@@ -0,0 +1,28 @@
+{
+    "name": "Homebridge",
+    "release": "11.2-RELEASE",
```
sonicaj: same ?

PR #7, opened 2019-11-06
Update README with more information

a-stjohn:

PR #8, opened 2019-11-11
Update README for iocage fetch

sonicaj:


sonicaj:
This will be fixed in https://github.com/ix-plugin-hub/iocage-plugin-index/pull/7

PR #11, opened 2019-11-12
Fix rtorrent-flood dependencies

pierrickrouxel:


miwi-fbsd:
Hi,

When you remove nat please set dhcp 1 to make sure the iocage gets the correct network setting.

Thanks.

pierrickrouxel:
It's ok thank you.

PR #26, opened 2020-01-20
Change GitHub artifact location

ConorBeh:
I changed my username to ConorBeh(my real name) and have updated the artifact location. I admittedly suck at using Git so this also contains an addition of PrivateBin. Sorry about the wacky PR's. 

You will notice that the server uses a self-signed SSL certificate. I discovered (with the help of Dan Langile) that PrivateBin cannot be accessed via Chrome/Chromium unless SSL is enabled. For this purpose I generated a junk self signed certificate that will expire in 10 years. I'm sure there are better methods for accomplishing this but this seems to work perfectly fine and thankfully lighttpd makes it easy to enable SSL.

### DIFF
--- a/privatebin.json
+++ b/privatebin.json
@@ -1,0 +1,26 @@
+{
+    "name": "PrivateBin",
+    "plugin_schema": "2",
+    "release": "11.3-RELEASE",
+    "artifact": "https://github.com/ConorBeh/iocage-plugin-privatebin.git",
+    "pkgs": [
+        "privatebin"
+        "lighttpd",
+        "ca_root_nss",
+        "openssl"
+    ],
+    "properties": {
+        "dhcp": 1
+    },
+    "packagesite": "http://pkg.FreeBSD.org/FreeBSD:11:amd64/latest",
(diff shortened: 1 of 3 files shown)


ConorBeh:
@miwi-fbsd When accessing PrivateBin over HTTP, the page will load but PrivateBin will not let the user to use the site. This is issue is sadly not fixable due to how the crypto is handled. Here's a screenshot of what happens (Chrome on Android): https://i.imgur.com/mZOOJfX.jpg

If you want to see what should happen (the only difference being SSL is enabled): paste.conorbeh.com

You are correct, I will remove the redundant :443 in ui.json

privatebin.json
```diff
@@ -0,0 +1,26 @@
+{
+    "name": "PrivateBin",
+    "plugin_schema": "2",
+    "release": "11.3-RELEASE",
+    "artifact": "https://github.com/ConorBeh/iocage-plugin-privatebin.git",
+    "pkgs": [
+        "privatebin"
```
miwi-fbsd: here is a , missing

PR #28, opened 2020-01-24
Create index.schema

kmoore134:


kmoore134:
You are right, didn't catch that it wasn't a comment. Updated

PR #40, opened 2020-02-03
Use this one! Port changes

overkillthemighty:
Made a mistake on the last port previously, use this one.


overkillthemighty:
The problem with that is that the WAPs default connect to the controller
over port 8080, I am not sure if there is a non-PITA method of changing
that.
If this is the case We probably need to make the controller have its own IP
and not be NATted.

PR #48, opened 2020-02-11
Sync with Master

kmoore134:
Sync a variety of new plugins with master and make live on 11.3.


INDEX
```diff
@@ -23,6 +23,22 @@
...
+    "description": "Duplicati 2.0 Store securely encrypted backups on cloud storage services!",
+    "official": false,
+    "icon": "https://icons.freenas.org/community-icons/duplicati.png",
+    "primary_pkg": "",
+    "category": "backup"
+  },
+  "esphome": {
+        "MANIFEST": "esphome.json",
+        "name": "ESPHome",
+        "icon": "https://raw.githubusercontent.com/tprelog/freenas-plugin-index/11.3-RELEASE/icons/esphome.png",
```
sonicaj: We need a `primary_pkg` key here, it could be null if there's no primary pkg. It is to keep consistent with all entries.

INDEX
```diff
@@ -79,7 +103,7 @@
   },
   "node-red": {
     "MANIFEST": "node-red.json",
-    "description": "Low-code programming for event-driven applications.",
+    "description": "Node-RED is a programming tool for wiring together hardware devices, APIs and online services in new and interesting ways.",
     "icon": "https://icons.freenas.org/community-icons/node-red.png",
     "name": "Node-RED",
     "official": false
```
sonicaj: We should have one for this plugin as well. Please let me know if you think i should perhaps do a clean up afterwards for all plugins ? Not sure how we have them in master :)

kmoore134:
> There's another thing, umm how are we handling packagesites ? Earlier we used to have 11.X packagesites separate so that the dependencies remain good. Now we have some plugins using unstable here, which earlier was development and it's possible whenever we update pkgs there, because of some dependencies, we break existing plugins in 11.3 ?

Thats a good question. I'm not sure if we standardize on using the public FreeBSD package sites, so we don't have to worry about breaking community plugins when we push changes to our testing packages? @william-gr any thoughts on this?

william-gr:
For community plugins I am not sure what else we could do.

Perhaps use could try to enforce the use of the quarterly package site so its not updated as often?

I dont want us to have the burden to maintain any packagesite for community

kmoore134:
At the moment I've let it slide, since no matter which repo is specified, the community contributor will have to monitor and track it to ensure they fix breakage introduced by upstream. Suggest we continue this for now, but perhaps for future submissions we start asking users to default to freebsd public repos, or other non-iX ones?

tprelog:
I'm a guilty party here. I will change mosquitto to quarterly and submit a pull request for sure.

For homeassistant and esphome I have a question about this change. 
Those services are both `python -m venv` then `pip3 install`

On using the latest package these are using `python=3.7.6`
On quarterly it is using `python=3.7.5`

How would changing the package sites affect someone if they installed the plugin using latest packages and then runs a plugin update that uses the quarterly packages?

sonicaj:
> I'm a guilty party here. I will at least change mosquitto to quarterly and submit a pull request.
> 
> EDIT: Looking at other plugin.json I see some are using
> 
> ```
> "packagesite": "http://pkg.cdn.trueos.org/iocage/11.3-RELEASE
> ```
> 
> Is the preferred over using
> 
> ```
>   "packagesite": "http://pkg.FreeBSD.org/${ABI}/quarterly",
> ```
> 
> For homeassistant and esphome I have a question about this change.
> Those services are both `python -m venv` then `pip3 install`
> 
> On using the latest package these are using `python=3.7.6`
> On quarterly it is using `python=3.7.5`
> 
> How would changing the package sites affect someone if they installed the plugin using latest packages and then runs a plugin update that uses the quarterly packages?

@tprelog that's a good question. So how iocage handles this is that it completely removes all packages from the container on update and then installs them new. The database/configuration files in place remain untouched so user settings are preserved this way. There shouldn't be issues ideally but there is no way to know for sure unless this is tried.

I would recommend installing the plugin with current packagesite and then updating the plugin manifest and then try plugin update. If that works as desired, the next step would be to make sure a fresh install works nicely as well - it should ideally but better to be sure about it.

tprelog:
> I would recommend installing the plugin with current packagesite and then updating the plugin manifest and then try plugin update.

Got it! From the point of maintaining this as a FreeNAS plugin, should I use
```
"packagesite": "http://pkg.cdn.trueos.org/iocage/11.3-RELEASE
```
I really want my plugins to be solid -- 

Thanks

sonicaj:
> > I would recommend installing the plugin with current packagesite and then updating the plugin manifest and then try plugin update.
> 
> Got it! From the point of maintaining this as a FreeNAS plugin, should I use
> 
> ```
> "packagesite": "http://pkg.cdn.trueos.org/iocage/11.3-RELEASE
> ```
> 
> I really want my plugins to be solid --
> 
> Thanks

Thank you @tprelog for your interest. So no, please use FreeBSD quarterly packagesite because the one you mentioned is specific to official FN plugins where we could update pkgs based on other plugin's dependencies which could potentially breaking yours.

PR #52, opened 2020-02-16
Initial commit of netdata plugin for community plugins. Light testing

kmoore134:
so far, but appears that disk,memory,cpu and similar charts are all working
properly


netdata.json
```diff
@@ -0,0 +1,33 @@
...
+    "mount_fdescfs": 1,
+    "mount_procfs": 1,
+    "securelevel": -1,
+    "allow_set_hostname": 1,
+    "allow_raw_sockets": 1,
+    "allow_sysvipc": 1,
+    "sysvmsg": "new",
+    "sysvsem": "new",
+    "sysvshm": "new",
+    "dhcp": "1"
```
sonicaj: Do you think we should default to NAT instead and add `nat_forwards` here ?

INDEX
```diff
@@ -102,6 +102,14 @@
     "official": false,
     "primary_pkg": "mosquitto"
   },
+  "netdata": {
+    "MANIFEST": "netdata.json",
+    "description": "Monitor everything in real-time",
+    "icon": "https://icons.freenas.org/community-icons/netdata.png",
+    "primary_pkg": "netdata",
+    "name": "Netdata",
```
sonicaj: We have a property `maintainer` as well which is used to track who is the maintainer of a plugin and users can reach out to the relevant maintainer if there is any issue with his/her plugin. Do you think we should have a pass at all the plugins here and add relevant maintainers if any ? ( Having a pass at all the plugins is of course not related to this PR, just some thoughts )

mgittelman:
Thanks for creating this!  Is there any way to get the network chart to work?  It only shows the bridge interface for the plugin, not for the FreeNAS server interfaces.

PR #61, opened 2020-04-09
Jackett plugin

fulder:
[Jackett](https://github.com/Jackett/Jackett) is an API wrapper for different trackers and can be used to generate feed URLs for Radarr, Sonarr and CouchPotato and this way used as an indexer for these services. It can also generate pure RSS streams.

This plugin will use the already present `jackett` package from the latest freebsd mirror. Except enabling and starting up the service the `post_install.sh` script will also make the `jackett` user the owner of the `/usr/local/share/jackett` folder in order to make the auto updates possible.

I've seen that this plugin was already added in tha past (https://github.com/ix-plugin-hub/iocage-plugin-index/search?q=jackett&type=Commits) but for some reason also removed a few days later? Was there any good reason for that removal? (Maybe the update bug?)


kmoore134:
Merged! I'll backport this to the 11.3 branch shortly :)

PR #63, opened 2020-04-11
Add HOOBS plugin

rahul-r:


hoobs.json
```diff
@@ -0,0 +1,30 @@
+{
+	"name": "hoobs",
+	"release": "11.3-RELEASE",
+	"artifact": "https://github.com/rahul-r/iocage-plugin-hoobs.git",
+	"properties": {
```
ConorBeh, line 5: Consider letting the user determine if they want the plugin to start at boot

hoobs.json
```diff
@@ -0,0 +1,30 @@
...
+		"boot": "on"
+	},
+	"pkgs": [
```
ConorBeh, line 8: Fix the indentation in the list of pkgs

hoobs.json
```diff
@@ -0,0 +1,30 @@
...
+		"node",
+		"npm",
+		"dbus",
+		"avahi-libdns",
+		"gcc",
+		"nss_mdns",
+		"jq"
+	],
+	"packagesite": "http://pkg.FreeBSD.org/FreeBSD:11:amd64/quarterly",
+	"fingerprints": {
```
ConorBeh, line 21: Fix the indents in this section

hoobs.json
```diff
@@ -0,0 +1,30 @@
...
+		"plugin-default": [
```
ConorBeh, line 22: Fix the indentation in this section

hoobs.json
```diff
@@ -0,0 +1,27 @@
+{
+    "name": "hoobs",
+    "release": "11.3-RELEASE",
+    "artifact": "https://github.com/rahul-r/iocage-plugin-hoobs.git",
+    "pkgs": [
+        "python",
+        "perl5",
+        "sudo",
```
fulder, line 8: Is `sudo` really needed? Can't find any usage of it in `post_install.sh`. I've tried to install the plugin without it and that worked good as well.

rahul-r: `sudo` is not needed for `post_install.sh` but used by the _HOOBS_ application. Without it, the application will fail to start.

This is the error message without `sudo`
```node
⠹ Enviornment: Clearing NPM Cacheevents.js:293
      throw er; // Unhandled 'error' event
      ^

Error: spawn sudo ENOENT
```

kmoore134:
Not sure if this is important, but saw these errors scroll by during the installation:

`Patching file /root/.hoobs/node_modules/@hoobs/systeminfo/lib/cpu.js using Plan A...
Hunk #1 succeeded at 668.
- Application
done
- Application
done
jq: error: Could not open file /root/.hoobs/etc/config.json: No such file or directory
done
done
rm: /root/.hoobs/etc/config.json: No such file or directory
done
done
mv: rename /tmp/config_tmp.json to /root/.hoobs/etc/config.json: No such file or directory
done`

Also, when finished, the IP it reports is missing the :8080 port on the end, so trying to launch the HOOBS UI fails until I manually entered that port number.

rahul-r:
Port 80 did not work because the step that sets the default port failed. I couldn't reproduce the error; I assume the configuration file 'config.json' doesn't get created on all systems. I have updated the post_install.sh script to create the configuration file from the template. Hopefully this fill fix the issue.
[here's the fix](https://github.com/rahul-r/iocage-plugin-hoobs/commit/923903a5bab94682854e8ef533f8bc81366c2b87)

rahul-r:
Sorry, I accidentally closed it

ConorBeh:
If you're not using port 80 for administration, make sure to update `ui.json` to point to the new port.

rahul-r:
I am using port 80. The issue Kris observed was because the script that sets the port to 80 failed. I’ve fixed the issue and the admin page should be accessible at 80

ConorBeh:
Ah, ok. My mistake.

PR #69, opened 2020-04-20
Update11.3 20200420

kmoore134:


INDEX
```diff
@@ -87,6 +103,13 @@
     "name": "Homebridge",
     "offical": false
   },
+  "hoobs": {
+    "MANIFEST": "hoobs.json",
+    "description": "HOOBS a Homebridge stack with a interface that simplifies configuration and installing plugins. This adds a process wrapper for Homebridge. It also adds an API that can be used as an endpoint for an application.",
+    "icon": "https://icons.freenas.org/community-icons/hoobs.png",
+    "name": "HOOBS",
+    "offical": false
```
sonicaj, line 111: We are missing `primary_pkg` attribute here, it can be `null`. Also I think we should default to `null` in the first one as well ( we try to do that everywhere where primary pkg has not been specified )

hoobs.json
```diff
@@ -0,0 +1,27 @@
...
+        "sudo",
+        "node",
+        "npm",
+        "dbus",
+        "avahi-libdns",
+        "gcc",
+        "nss_mdns",
+        "jq"
+    ],
+    "packagesite": "http://pkg.FreeBSD.org/FreeBSD:11:amd64/quarterly",
```
sonicaj, line 17: Should we use `ABI` here ?

PR #73, opened 2020-05-17
Add calibre-web plugin

aserrallerios:


calibre-web.json
```diff
@@ -0,0 +1,26 @@
...
+    ],
+    "properties": {
+        "dhcp": 1
+    },
+    "packagesite": "http://pkg.FreeBSD.org/${ABI}/latest",
+    "fingerprints": {
+        "plugin-default": [
+            {
+                "function": "sha256",
+                "fingerprint": "226efd3a126fb86e71d60a37353d17f57af816d1c7ecad0623c21f0bf73eb0c7" // ???
```
fulder, line 20:
```suggestion
                "fingerprint": "b0170035af3acc5f3f3ae1859dc717101b4e6c1d0a794ad554928ca0cbb2f438"
```

Should probably be the correct trusted fingerprint (you can check the other plugins have the same one for the latest packagesite)

calibre-web.json
```diff
@@ -0,0 +1,26 @@
+{
+    "name": "Calibre-Web",
+    "plugin_schema": "2",
+    "release": "11.3-RELEASE",
+    "artifact": "https://github.com/entuerto/iocage-plugin-calibre-web.git",
+    "pkgs": [
+        "python37",
+        "py37-pip",
+        "py37-setuptools",
+        "py37-sqlite3"
```
fulder, line 10: Saw you are using `git` in your `post_install.sh`. Missing here?

INDEX
```diff
@@ -7,6 +7,14 @@
     "official": false,
     "primary_pkg": null
   },
+  "calibre-web": {
+    "MANIFEST": "calibre-web.json",
+    "description": "Web app for browsing, reading and downloading eBooks stored in a Calibre database",
+    "icon": "https://icons.freenas.org/community-icons/calibre-web.png",
+    "name": "Calibre-Web",
+    "official": false,
+    "primary_pkg": ""
```
fulder, line 16:
```suggestion
    "primary_pkg": null
```

aserrallerios:
The author of the artifact repository seems to be not available at the moment. I wonder if we should clone it? Or it's ok?

kmoore134:
If you would like to maintain the repo, then yes it is possible to fork it. I saw that on the original repo it hasn't been touched in 5 months, and your open issue is about 2 weeks old, so that seems like a reasonable timeout.

aserrallerios:
Done, I think it's fair to assume that should it raise any problem with the plugin, I'll be more reachable than the original author.

kmoore134:
Getting closer to working. Tested here, and it looks like the service fails to start due to missing python3 symlink or binary:

`root@Calibre-Web:/usr/local/etc/rc.d # service calibre_web start
Starting calibre_web
root@Calibre-Web:/usr/local/etc/rc.d # daemon: python3: No such file or directory`

aserrallerios:
Should be working now.

Sorry for my ignorance, it's not 100% clear to me how to test it end-to-end. I guess it should be something similar to...

iocage fetch -P --name calibre-web ip4_addr="xxxx" --branch 'master'

kmoore134:
I'm testing with this:

`iocage fetch -P ./calibre-web.json dhcp=1`

It gets farther, but fails here now:

`Cloning into '/usr/local/app/calibre-web'...
Traceback (most recent call last):
  File "/usr/local/app/calibre-web/cps.py", line 34, in <module>
    from cps import create_app
  File "/usr/local/app/calibre-web/cps/__init__.py", line 28, in <module>
    from babel import Locale as LC
ModuleNotFoundError: No module named 'babel'
Cloning into '/usr/local/app/calibre-web'...

Admin Portal:
http://10.238.11.210:8083/
`

fulder:
> I'm testing with this:
> 
> `iocage fetch -P ./calibre-web.json dhcp=1`
> 
> It gets farther, but fails here now:
> 
> ```
> Cloning into '/usr/local/app/calibre-web'...
> Traceback (most recent call last):
>   File "/usr/local/app/calibre-web/cps.py", line 34, in <module>
>     from cps import create_app
>   File "/usr/local/app/calibre-web/cps/__init__.py", line 28, in <module>
>     from babel import Locale as LC
> ModuleNotFoundError: No module named 'babel'
> Cloning into '/usr/local/app/calibre-web'...
> 
> Admin Portal:
> http://10.238.11.210:8083/
> ```

@aserrallerios if it is of any help I think this line in your `post_install.sh` could be the issue:

```
# Install dependencies 
pip install --target vendor -r /usr/local/app/calibre-web/requirements.txt
```

The `vendor` folder containing the python modules in installed in the `/` directory and not in `/usr/local/app/calibre-web` as the application expects.

Think it should be:
```
# Install dependencies 
pip install --target /usr/local/app/calibre-web/vendor -r /usr/local/app/calibre-web/requirements.txt
```
At least while testing fast I could start up the service and properly and the webUI worked after moving this folder.

aserrallerios:
It also works if I remove the `--target vendor`.

What do you prefer?

fulder:
> It also works if I remove the `--target vendor`.
> 
> What do you prefer?

Yes and just install the modules globally, that's probably more pretty while inside a iocage :+1:

fulder:
Sorry for a bit late reply but my FreeNAS PSU died and had to replace it :P 

It looks like the installation works fine for me now, so fine for me to merge this :+1:

PR #75, opened 2020-05-24
add openssl

tprelog:
The HomeKit integration and possibly others, require openssl-1.1.1

### DIFF
--- a/homeassistant.json
+++ b/homeassistant.json
@@ -16,6 +16,7 @@
     "gmake",
     "libxml2",
     "libxslt",
+    "openssl",
     "pkgconf",
     "python37",
     "py37-av",


kmoore134:
Have you tested this yet? Normally openssl comes from the base system, or is automatically installed as a dependency.

tprelog:
Hi @kmoore134, so I've tested some changes locally, as much as I can it's working. I don't own any physical HomeKit devices so I am relying on user feedback. [This solution looks promising](https://github.com/tprelog/iocage-homeassistant/issues/14#issuecomment-633141287)

I have already updated the `post_install.sh` locally and Home Assistant appears to now be using `openssl-1.1.1` --  I am still waiting on more confirmation before I push my changes to the artifact repository.

fulder:
I've run a fast test installing this plugin and then checking if openssl got installed. It wasn't added by any other dependency/OS so guess in order to fix the bug @tprelog linked this will be needed.

tprelog:
@fulder you got it! -- `openssl` in the 11.3-RELEASE base is too old!

If it would be helpful, I can provide several links this has been discussed related to Home Assistant.

Also Home Assistant is a Python application installed using `pip` - It does not install or update any system dependencies on it's own.

From the [Home Assistant docs for HomeKit](https://www.home-assistant.io/integrations/homekit/)
>HomeKit requires openssl 1.1.0 or later as the HomeKit Accessory Protocol (HAP) uses the ChaCha20 stream cipher and the Poly1305 authenticator.

As of today, Home Assistant supports [over 1600 integrations](https://www.home-assistant.io/integrations/)  I would be surprised if HomeKit was the only one needing openssl-1.1.1

PR #78, opened 2020-05-31
Grafana (with InfluxDB)

MalteHillmann:
Hello,

I've created some simple Grafana-Plugin which installs Grafana and InfluxDB with a basic setup (with basic auth).

Requirements are met:
1. Added the entry to INDEX in alphabetical order 
2. Included a new icon in the ./icon/ directory
3. Submitted icon file is a valid PNG that is 128x128 in size

Kind regards
Malte Hillmann


INDEX
```diff
@@ -96,6 +96,14 @@
     "official": false,
     "primary_pkg": "gogs"
   },
+  "grafana": {
+    "MANIFEST": "grafana.json",
+    "description": "Grafana allows you to query, visualize, alert on and understand your metrics no matter where they are stored. Create, explore, and share dashboards with your team and foster a data driven culture.",
+    "icon": "https://icons.freenas.org/community-icons/grafana.png",
+    "name": "Grafana (with InfluxDB)",
+    "official": false,
+    "primary_pkg": "grafana"
```
kmoore134, line 105: Shouldn't this be grafana6 to match whats installed in the package list in grafana.json?

grafana.json
```diff
@@ -0,0 +1,25 @@
...
+    "artifact": "https://github.com/MalteHillmann/iocage-plugin-grafana.git",
+    "pkgs": [
+        "grafana6",
+        "influxdb",
+        "curl"
+    ],
+    "properties": {
+        "dhcp": 1
+    },
+    "packagesite": "http://pkg.FreeBSD.org/FreeBSD:11:amd64/latest",
```
kmoore134, line 14: Can you use "http://pkg.FreeBSD.org/${ABI}/latest" instead so that it works on upcoming 12.0 as well?

MalteHillmann:
Hi @kmoore134 , you are right. I changed both lines according to your suggestions.

fulder:
Good job with this plugin :) I have just a few comments when it comes to the `post_install.sh` script.

I'm not a big fan with sleeps in scripts. What if the InfulxDB and/or grafana service start up slow? 

For InfulxDB you could do `curl -sL -I localhost:8086/ping?wait_for_leader=30s` (source: https://github.com/influxdata/influxdb/issues/4595#issuecomment-151958860)
and for grafana something like .e.g ` [ "$(curl -I -s -o /dev/null -w 200 http://localhost:3000/api/health)" == "200" ]; then echo "aa"; fi`

fulder:
Good job with this plugin :) Tested and it seems to work fine.

I have just a few comments on your `post_install.sh` script:

---
I'm not a big fan with random sleeps in scripts. What if the InfulxDB and/or grafana service start up slow?
For InfulxDB you could do: 
```
curl -sL -I localhost:8086/ping?wait_for_leader=30s
``` 
(source: https://github.com/influxdata/influxdb/issues/4595#issuecomment-151958860)
and for grafana something like .e.g 
```

while [ "$(curl -I -s -o /dev/null -w 200 http://localhost:3000/api/health)" != "200" ]; do sleep 1; done
```
----
It's good you create a new user with a random password for the InfluxDB user but is it really good to change the default grafana admin password to something random as well? Can't this be done by through the UI after a first login instead?

fulder:
Although Grafana seems to be up and running and I managed to log in I noticed some error/warnings logs in `/var/logs/grafana/grafana.log`

Not sure if this is anything serious but I got this prints (event during a stop/start of the service after waiting a while post the installation):

```
t=2020-05-31T23:19:47+0200 lvl=eror msg="can't read datasource provisioning files from directory" logger=provisioning.datasources path=/var/db/grafana/provisioning/datasources error="open /var/db/grafana/provisioning/datasources: no such file or directory"
t=2020-05-31T23:19:47+0200 lvl=eror msg="Can't read alert notification provisioning files from directory" logger=provisioning.notifiers path=/var/db/grafana/provisioning/notifiers error="open /var/db/grafana/provisioning/notifiers: no such file or directory"
t=2020-05-31T23:19:47+0200 lvl=eror msg="can't read dashboard provisioning files from directory" logger=provisioning.dashboard path=/var/db/grafana/provisioning/dashboards error="open /var/db/grafana/provisioning/dashboards: no such file or directory"
t=2020-05-31T23:19:47+0200 lvl=info msg="Backend rendering via phantomJS" logger=rendering renderer=phantomJS
t=2020-05-31T23:19:47+0200 lvl=warn msg="phantomJS is deprecated and will be removed in a future release. You should consider migrating from phantomJS to grafana-image-renderer plugin. Read more at https://grafana.com/docs/grafana/latest/administration/image_rendering/" logger=rendering renderer=phantomJS
```

MalteHillmann:
Hi @fulder,

thank you for your feedback. 
I added your proposed health checks instead of that 5s "random" sleeps. Also the directory creations are in the script now.

> It's good you create a new user with a random password for the InfluxDB user but is it really good to change the default grafana admin password to something random as well (took me a while before I realized it wasn't the default password after installation was complete)? Can't this be done through the UI after a first login instead?

Yes, password change in Grafana UI is possible and forced on first login. But I personally don't like autoinstall-scripts with default passwords. Because if there is time between install and first login, the grafana instance can possibly be compromised. (I know, the chance is very small, but it's there, I think)

Kind regards
Malte Hillmann

fulder:
Nice job with the quick fixes @MalteHillmann, looks like it still works fine and the errors in log are gone. I understand your worry about compromised time, setting a random password does lower the risk. So fine to keep it that way. =)

Can approve this one

PR #81, opened 2020-06-07
Gitlab plugin

pandorixtech:
This will allow to install gitlab as a plugin using dhcp.


gitlab.json
```diff
@@ -0,0 +1,27 @@
...
+  "packagesite": "http://pkg.FreeBSD.org/${ABI}/latest",
+  "fingerprints": {
+	  "iocage-plugins": [
+		  {
+		  "function": "sha256",
+                  "fingerprint": "b0170035af3acc5f3f3ae1859dc717101b4e6c1d0a794ad554928ca0cbb2f438"
+	  }
+	  ]
+  },
+  "official": true
```
fulder, line 26: Shouldn't this be `"official": false` now that it has been moved to this community plugin index (and to match the INDEX official value)?

pandorixtech: Correct! Thank you for your review, I've just pushed a fix for it.

gitlab.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "GitLab",
+  "release": "11.3-RELEASE",
+  "artifact": "https://github.com/pandorixtech/iocage-plugin-gitlab.git",
+  "pkgs": [
+    "www/gitlab-ce",
```
fulder, line 6: Nothing crucial, but wouldn't it be more consistent to either use the category prefix for all package or for none? I.e. just `gitlab-ce` works fine here too.

pandorixtech: No problem! :) 
category prefix removed.
Thank you for your review.

fulder:
Got some errors/warnings prints during the installation of the plugin (nothing really serious, see it more as testing from my side and not a "must fix", feel free to resolve):

---

Postgres warnings:

```
/root/post_install.sh: cannot create /usr/local/pgsql/data/postgresql.conf: No such file or directory

/root/post_install.sh: cannot create /usr/local/pgsql/data/pg_hba.conf: No such file or directory
```

Probably nothing serious as the server seems to start up anyway.

---

```
rm: Gemfile.lock: No such file or directory
```

Could probably be solved by adding a `-f` flag to the Gemfile lock removal

---
Nothing to fix right away (could maybe be done during plugin optimization and has more to do with gitlab implementation than plugin implementation) but the installation was a bit slow (and took a lot of my CPU).

The locale settings and precompile was a bit slow:
`'rake:assets:precompile' finished in 117.112188414 seconds`

but this webpack command seemed to get stuck for a very long time:
`/usr/local/www/gitlab-ce/node_modules/.bin/webpack --config /usr/local/www/gitlab-ce/config/webpack.config.js --bail`

`14031 git            11  96    0  2052M  1913M RUN     0  10:10 108.17% node`

```
Hash: f8c3a57250044a1129cf
Version: webpack 4.42.0
Time: 674559ms
```
(that's around 11m for that task)

---

After this the installation seems to start up gitlab properly :+1: I've tried to create a user, upload a SSH key, create a repo, clone it and push to it. Seems fine

gitlab.json
fulder: Cool, thanks for fast fix =)

pandorixtech:
@fulder 
Thank you for your review.
The post_config unnecessary commands (due to postgresql version update) were removed, since it will make it cleaner.
:)

pandorixtech:
@kmoore134 
Thank you for your review, the icon has been resized to 128x128.

PR #83, opened 2020-06-11
Introduce Heimdall plugin

jsegaert:

- [x] 1. Add the entry to INDEX in alphabetical order  
- [x] 2. Includes a new icon in the ./icon/ directory
- [x] 3, Ensure the submitted icon file is a valid PNG that is 128x128 in size


INDEX
```diff
@@ -112,6 +112,14 @@
     "official": false,
     "primary_pkg": "grafana6"
   },
+  "heimdall-dashboard": {
+    "MANIFEST": "heimdall-dashboard.json",
+    "description": "Heimdall Application Dashboard is an elegant solution to organize all your web applications and links onto a single page website.",
+    "icon": "https://icons.freenas.org/community-icons/heimdall-dashboard.png",
+    "name": "Heimdall",
+    "official": false,
+    "primary_pkg": "Git branch - Master"
```
fulder, line 121: How does this primary_pkg with a custom string work? Shouldn't it be `null` in this case?

kmoore134: Correct, if this isn't a legit package name, it should be set to "null"

jsegaert: Thank you for the feedback; I'll fix this.

PR #87, opened 2020-06-25
update homeassistant for TrueNAS Core

tprelog:
Home Assistant _Core_ on TrueNAS _Core_...

getting to the _Core_...
it's all about the _Core_...
this release will be hard _Core_..

:rofl:  :roll_eyes:

Issue #80, opened 2020-06-05
Need solution when plugins require access to USB devices


tprelog:
The Home Assistant Core [community plugin](https://github.com/tprelog/iocage-homeassistant) currently requires a custom devfs_ruleset in order to access USB Z-Wave and ZigBee controllers. On FreeNAS, these devices show as `/dev/cua*/` -- For example the [HUSBZB-1](https://shop.homeseer.com/products/nortek-usb-zigbee-zwave-interface) contains both Z-Wave `/dev/cuaU0` and ZigBee `/dev/cuaU1` radios
 
The current approach requires requires additional user configuration
- [Using a USB Z-Wave or ZigBee controller][1]

[1]: https://github.com/tprelog/iocage-homeassistant/wiki/Using-a-USB-Z-Wave-or-Zigbee-controller

_Is there a better solution to accomplish this?_

_Can it be made possible, to automatically pass through external USB devices when required ?_

The current approach [also has an issue causing the ruleset to get deleted](https://github.com/tprelog/iocage-homeassistant/issues/11) when restarting the jail using the FreeNAS web-ui. The only work around I have found is using the console to preform a "soft restart"
```
iocage restart -s homeassistant
```

---

I have also considered creating additional FreeNAS Community Plugins for [ozwdaemon](https://github.com/OpenZWave/qt-openzwave#home-assistant-mqtt-client-adapter) and [zigbee2mqtt](https://github.com/Koenkk/zigbee2mqtt#zigbee2mqtt---), both of which would require the same type of device pass through.

jsegaert:
There's a `devfs_ruleset` option that was implemented many moons ago for this purpose.
The idea is to specify the custom ruleset in the plugin's manifest file, e.g. something like:

```
  "devfs_ruleset": {
      "paths": {
          "zfs": "unhide",
          "bpf*": "unhide",
          "cua*": "unhide"
       },
      "includes": [
          "1",
          "2",
          "3"  ]
      },
```
I haven't come across any plugin that is actually using it, nor have I played with it myself, but this feature was requested by @kmoore134 himself.

Source: https://redmine.ixsystems.com/issues/42593

tprelog:
[...]

PR #89, opened 2020-06-27
update homeassistant pkgs

tprelog:
The current solution to use openssl-1.1.1 for HomeKit, will break Z-Wave for everyone using the 11.3-RELEASE of this plugin 
(even non HomeKit users) - I think it is unacceptable to by default, apply a workaround or update something in this plugin, that will break something else, that should otherwise work if you followed a manual installation. For this reason, the updated openssl should no longer be included in this plugin by default.

HomeKit users can still choose to manually update openssl at any time. - However HomeKit users wishing to also use Z-Wave, will require a second workaround to get Z-Wave working again.


tprelog:
Hi @sonicaj --  I'm still trying to figure out how to apply this correctly for the 11.3-RELEASE however this is already non-issue for TrueNAS Core 12 since the base already ships with an updated version of openssl, meeting the requirements for HomeKit.

The issue in 11.3-RELEASE is not much directly because of `openssl` but rather a side effect of the way it is currently being applied to Home Assistant. ( I think the problem is with `libiconv` )

Trying to keep this brief -- `ffmpeg` is required by multiple components, removing this will cause build dependency issues for largest number of people... `ffmpeg` is installing `libiconv`

So here's how that come into play. Keep in mind Home Assistant is a Python application, installed and updated using pip in a virtualenv. Basically in order to install Home Assistant so it is using openssl-1.1.1, the current solution ( that breaks zwave ) is to set these variables before installing or updating HA
```
export PATH=/usr/local/sbin:/usr/local/bin:${PATH}
export CPATH=/usr/local/include
export LIBRARY_PATH=/usr/local/lib
```

The problem is that when `libiconv` is installed ( for ffmpeg ) and `CPATH=/usr/local/include` ( for openssl ) Z-Wave fails to install or update with the following error
```
ImportError: /srv/homeassistant/lib/python3.7/site-packages/libopenzwave.so: Undefined symbol “libiconv”
```

There are a few ways I've gotten past that

1. `pkg delete libiconv` before installing Home Assistant -- both HomeKit and Z-Wave work but what effect will this have on `ffmpeg` - This is the easiest solution but seems least correct

2. Manually reset variables and reinstall Z-Wave -- but this will need to be repeated with every update to Z-Wave. This effects every Z-Wave user, even ones not using HomeKit
```
export PATH=/usr/local/sbin:/usr/local/bin:${PATH}
export CPATH=
export LIBRARY_PATH=
pip install --force-reinstall --no-cache homeassistant-pyozw
```

At this time, I feel reverting the update to `openssl`  will better limit the effects of all this in the following way

1. Z-Wave will work for everyone again

2. HomeKit users can still update openssl manually ( This will be a one time manual intervention )
- Z-Wave issue is limited to smallest group possible ( Only those using both HomeKit and Z-Wave )
- HomeKit user _not using_ Z-Wave are not effected by second issue

tprelog:
On a side note related to this plugin, do you know if the changes in iocage that allow the devfs_rule set to work, will be back ported to the 11.3 release.

I have tested and it is working on TrueNAS core 12 nightlies but I am receiving a trace back error when I try adding it to this 11-3 manifest.

sonicaj:
@tprelog can you please confirm if you are referring to these props https://github.com/iocage/iocage/commit/49e6280f67fb800251daa5373001af43a4818893 ?

tprelog:
@sonicaj 

Yes
@jsegaert has helped me with this here
https://github.com/ix-plugin-hub/iocage-plugin-index/issues/80 

I have tested and confirmed this works as I expected TrueNAS Core 12
https://github.com/ix-plugin-hub/iocage-plugin-index/pull/87/commits/49cad60126a457e1509979b93c12c9a3c85e89be

sonicaj:
Thank you for confirming, can you please create an issue on https://jira.ixsystems.com ? We can then see if we can port these changes to 11.3. When the changes were introduced, it was considered invasive for 11.3 and we didn't backport them at that time - also we had no known user actually using them at the time, so that was a factor as well to not backport them at the time.

tprelog:
I made it short it sweet ( personally, I feel my bug / issue reporting, needs much improvement ) 

I can provided any additional information, if it is needed
https://jira.ixsystems.com/browse/NAS-106614

tprelog:
Do I need to do something else before this can get merged?

tprelog:
Perfect I was just going to ask about that, I'll update the 12.1-RELEASE manifest. Thank you

PR #90, opened 2020-06-27
Bind plugin

pandorixtech:
Plugin to install BIND v9 as DNS


INDEX
```diff
@@ -7,6 +7,14 @@
     "official": false,
     "primary_pkg": null
   },
+  "bind": {
+    "MANIFEST": "bind.json",
+    "description": "BIND DNS server",
+    "icon": "https://icons.freenas.org/community-icons/bind.png",
+    "name": "Bind DNS",
+    "official": false,
+    "primary_pkg": null
```
fulder, line 16: As you're using the official `bind916` package from the FreeBSD package mirror, you could probably set the `primary_pkg` to `bind916` as well.

bind.json
```diff
@@ -0,0 +1,25 @@
...
+    "artifact": "https://github.com/pandorixtech/iocage-plugin-bind.git",
+    "official": false,
+    "properties": {
+        "nat": 1,
+        "nat_forwards": "udp(53:53)"
+    },
+    "pkgs": [
+        "bind916",
+	"vim",
+	"mc"
```
fulder, line 13: Is `vim` and `mc` really required by the bind package? I couldn't see any usages of it in your `post_install.sh` either.

Also the JSON indent got a bit weird in this pkgs array :)

INDEX
pandorixtech: Thank you for your review, the primary_pkg has been set.

pandorixtech: also the requested files have been added to the overlay dir and creation removed from post_install.sh

bind.json
pandorixtech: Thank you for your review, the format has been corrected. 
Although vim and mc are not required to run bind, since the configuration is done by editing configuration files, I decided to add the most common tools used by administrators for that, vim editor and midnight commander. If that is an issue to accept this plugin I will remove them.

fulder: No problem, you can keep them here =) Was more just wondering if it was a requirement for something else.

INDEX
fulder: Nice, good job with fast fixes :+1:  

The `/usr/local/etc/namedb/dynamic.local.db` this file doesn't seem to exist by the way so the [chown command](https://github.com/pandorixtech/iocage-plugin-bind/blob/master/post_install.sh#L12) in `post_install.sh` will not have any effect. Shouldn't this file exist or maybe remove the chown if it doesn't?

pandorixtech: sorry about that , when deleting extra content from post_install.sh I ended up making a typo mistake. Fixed now.
Thank you for your review :)

PR #93, opened 2020-06-30
update homeassistant for TrueNAS Core

tprelog:
@sonicaj 
 I think... Is what you were asking for ?


tprelog:
Thank you -- I've been testing this plugin using the TN nightly for a few weeks now and I'm very excited for the Beta 1

PR #98, opened 2020-07-14
Add www/chromium dependency for channels-dvr

tmm1:
Replacement for https://github.com/freenas/iocage-ix-plugins/pull/166 which didn't get merged due to accidental close.

We have had to guide several users through manual jail connect + package install, so it would be very helpful to have this included by default.

### DIFF
--- a/channels-dvr.json
+++ b/channels-dvr.json
@@ -9,7 +9,8 @@
     "pkgs": [
         "ftp/curl",
         "security/ca_root_nss",
-        "security/sudo"
+        "security/sudo",
+        "www/chromium"
     ],
     "packagesite": "http://pkg.cdn.trueos.org/iocage/unstable",
     "fingerprints": {
--- a/channels-dvr.json
+++ b/channels-dvr.json
@@ -17,7 +17,7 @@
         "iocage-plugins": [
             {
                 "function": "sha256",
-                "fingerprint": "226efd3a126fb86e71d60a37353d17f57af816d1c7ecad0623c21f0bf73eb0c7"
+                "fingerprint": "b0170035af3acc5f3f3ae1859dc717101b4e6c1d0a794ad554928ca0cbb2f438"
             }
         ]
     },
--- a/channels-dvr.json
+++ b/channels-dvr.json
@@ -12,7 +12,7 @@
         "security/sudo",
         "www/chromium"
     ],
-    "packagesite": "http://pkg.cdn.trueos.org/iocage/unstable",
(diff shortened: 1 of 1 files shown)


tmm1:
> Note that most plugins have been changed to use the standard FreeBSD pkg repo; it looks like http://pkg.cdn.trueos.org/iocage/unstable hasn't been updated in a while and TrueOS development has been discontinued.

Thank you for pointing that out! I will update to match.

What is the significance of the `release` field? I'm not sure why it says 11.3. Is that just the version used in the jail? I can update that to 12.1 at the same time if that makes sense.

channels-dvr.json
```diff
@@ -9,9 +9,10 @@
     "pkgs": [
         "ftp/curl",
         "security/ca_root_nss",
-        "security/sudo"
+        "security/sudo",
+        "www/chromium"
     ],
-    "packagesite": "http://pkg.cdn.trueos.org/iocage/unstable",
+    "packagesite": "http://pkg.FreeBSD.org/${ABI}/latest",
```
jsegaert, line 15: If you change the "packagesite", you also need to change the value of "fingerprint" to b0170035af3acc5f3f3ae1859dc717101b4e6c1d0a794ad554928ca0cbb2f438

tmm1: Thank you! I was wondering what that checksum was for.

jsegaert:
> What is the significance of the release field? I'm not sure why it says 11.3. Is that just the version used in the jail? I can update that to 12.1 at the same time if that makes sense.

And yes, the "release" determines the FreeBSD version of the userland in the plugin jail.  At the moment, the productive version of FreeNAS is still based on 11.3, but TrueNAS CORE beta is out, which is based on 12.1, which is why some plugins already switched to 12.1.  The `master` branch is considered development, the actual changes won't be visible to the actual FreeNAS until they get merged in the branch that corresponds with the FreeNAS release, e.g. `11.3-RELEASE`.

tmm1:
Okay I really appreciate the guide.

Sounds like this can be merged, then I should open another PR against the release branch to backport?

Then later I can open another PR to bump the release on master.

jsegaert:
OK, I went through the installation again and with the repo switch, the version of chromium went from 78 to 83, woohoo!

Once it's merged into `master` I would request to merge it in the `11.3-RELEASE` branch (this is done periodically anyway) and then open a PR to bump to 12.1 later.

jsegaert:
@kmoore134  Any feedback on this PR?

PR #112, opened 2020-10-01
Add Vault

marblenix:
Adds hashicorp's vault service

1. Add the entry to INDEX in alphabetical order ✔️
2. Includes a new icon in the ./icon/ directory ✔️
3. Ensure the submitted icon file is a valid PNG that is 128x128 in size ✔️


INDEX
```diff
@@ -448,6 +448,16 @@
         "category": "content",
         "homepage": "http://qbittorrent.org/"
     },
+    "vault": {
+        "MANIFEST": "vault.json",
+        "name": "vault",
+        "icon": "https://icons.freenas.org/community-icons/vault.png",
+        "description": "Manage secrets and protect sensitive data.",
+        "official": false,
+        "primary_pkg": null,
```
kmoore134, line 457: This looks good, but did you intend to set primary_pkg: null? It could probably be set to vault, so in the UI it shows the version of the freebsd vault package?

marblenix: An easy change! One moment

kmoore134:
Oh, as a follow-up. Did you intend to use "release": "11.3-RELEASE", or will 12.1-RELEASE work for the upcoming 12.1-RELEASE packages?

marblenix:
I was only able to test 11.3-RELEASE locally because of my freenas version; but I believe it should work with 12.1 out of the box.

fulder:
I just tested running it inside a 12.1 jail in a TrueNAS Core system. Looks like the installation is working fine :) Just one small thing. The services is not automatically started after the installation, so maybe something to add to the `post_install.sh` script. After manually starting it, I could reach the UI without any problems.

fulder:
Also there seems to be a small bug in the `ui.json` as the `Admin Portals` value for this plugin became `http://192.168.1.193:"0.0.0.0:8200"/ui/` after the installation (probably because the port value in the `vault.json` becomes `0.0.0.0:8200`.

marblenix:
@fulder 
thanks for testing that for me! new code with bugfixes should be live.

Issue #107, opened 2020-09-12
Sick Chill plugin install fails on Stable Release channel [11.3-RELEASE]


vipero07:
This may be related to sick chill's removal of python 27 support.

```
Install
Error: sickchill had a failure Exception: RuntimeError Message: Last 10 lines: /root/post_install.sh: git: not found /root/post_install.sh: pip: not found chown: /usr/local/app/sickchill: No such file or directory cp: /usr/local/app/sickchill/contrib/runscripts/init.freebsd: No such file or directory chmod: /usr/local/etc/rc.d/sickchill: No such file or directory Partial plugin destroyed
```

fulder:
From what I can see there seems to be a fix for python upgrade for this which was merged to master just 3 days before this issue was created: https://github.com/ix-plugin-hub/iocage-plugin-index/pull/106. But this fix has only been merged back to the 12.1-RELEASE branch (at least following the PR history)

Although the error you are getting are probably because pip and git is not installed but the plugin has started to need them both after this PR: https://github.com/ix-plugin-hub/iocage-plugin-sickchill/pull/1.

So in order to fix this master could maybe be synced to 11.3-RELEASE if this is OK @kmoore134 ? Will running FreeNAS 11.3 support running plugins with newer FreeBSD versions or will this cause more issues?

Otherwise it could also by solved by creating a patch update for the 11.3-RELEASE branch only with the python update and git/pip installation (would probably be a good idea to lock the release branches in the SickChill plugin for different kind of releases instead of just having 1 master for all of them).

(A bit outside of this issue but I've tested installing the plugin on a 12.1-RELEASE TrueNAS Core system and it worked fine)

turboture:
Hi
Is it possible for you to help us with a updated plugin or a patch?

turboture:
@fulder 
Is it possible for you to help us?

fulder:
Could this PR maybe solve the issue: https://github.com/ix-plugin-hub/iocage-plugin-index/pull/134 ?

I don't have a 11.3-RELEASE system to test with but if you could check you could probably try to install the plugin manually like this:

* `fetch https://raw.githubusercontent.com/fulder/iocage-plugin-index/11.3-RELEASE/sickchill.json`
* `iocage fetch -P sickchill.json --name sickchill_test dhcp=on`

turboture:
@fulder  Thank you for trying to help us. I got these errors. I tried to install it from ssh and shell. Same result.

[...]

PR #136, opened 2020-12-11
Add SickChill missing libxslt pkg needed for pip lxml

fulder:
Add missing SickChill pkg making the requirements installation in `post_install.sh` log error messages, see e.g.: https://github.com/ix-plugin-hub/iocage-plugin-index/issues/107#issuecomment-743286926

### DIFF
--- a/sickchill.json
+++ b/sickchill.json
@@ -13,7 +13,8 @@
         "unrar",
         "ca_root_nss",
         "py37-pillow",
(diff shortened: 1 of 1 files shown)


kmoore134 (approved):

fulder:
Thanks for the merge, could this be ported back to 11.3-RELEASE as well when it's time?

PR #149, opened 2021-01-19
Update INDEX with Dnsmasq svg logo

justinclift:
Note that we purposely leave the .png version of the logo in place, just in case it's still needed.

[x] - Add the entry to INDEX in alphabetical order
[x] - Includes a new icon in the ./icon/ directory
[ ] - Ensure the submitted icon file is a valid PNG that is 128x128 in size
  * **Nope**.  This commit switches to an svg version of the logo, which has been manually resized to 128x128

As an aside, is there a reasonably easy way for plugin developers to check how their logo will look in the list, prior to having their PR be merged?  eg without needing to setup a full FreeNAS build chain and point it at an alternative INDEX location :wink:

Looking over the README, the examples so far are more about testing updated plugin .json and script files.  Seems to slightly miss this (edge?) case of adjusting a logo. :grin:


kmoore134 (approved):

kmoore134:
Thanks! Right now the way to test updating JSON / icons locally is to go edit the files in this directory:

/mnt/<poolname>/iocage/.plugins/github_com_ix-plugin-hub_iocage-plugin-index_git

kmoore134:
FYI - I also tested the switch to SVG this way, and it appears to work fine :)

justinclift:
> FYI - I also tested the switch to SVG this way, and it appears to work fine :)

Awesome! :smile:

Now that it's merged, I just tried the "Refresh Index" button in the FreeNAS 11-3.U5 UI, but it's still pulling down the .png.  Guess it might need a bit of time, or I'm just too sleepy to realise something obvious.  I'll try it again in a few hours, after I've gotten some sleep.

kmoore134:
Ahh, right now its only backported to 12 branch, we tend to not want to touch 11.3 :) When you update to 12 later it should show right up.

justinclift:
Cool, that makes good sense.  "Don't mess with the stable thing...." approach.

Issue #165, opened 2021-02-17
Mono dependent plugins .e.g radarr and potentially sonarr no longer work


balbassam:
It looks like to me that something broke mono in the latest updates.

Steps to reproduce on TrueNAS are as follows:

Install radarr (or sonarr) plugin using the gui. Find that the management interface doesn't work. 

Running a shell in the plugins jail, I tried to run the binaries directly and find that there's an undefined symbol in mono.

For sonarr:
```
/usr/local/bin/mono /usr/local/share/sonarr/NzbDrone.exe --nobrowser --data=/usr/local/sonarr
ld-elf.so.1: /usr/local/bin/mono: Undefined symbol "pthread_setname_np@FBSD_1.6"
```

For radarr:
```
/usr/local/bin/mono /usr/local/share/radarr/Radarr.exe
ld-elf.so.1: /usr/local/bin/mono: Undefined symbol "pthread_setname_np@FBSD_1.6"
```

I have a sonarr plugin that I haven't updated that still works, while my radarr plugin that I did update recently broke.

repvik:
Can confirm this issue. Seeing the same missing symbol.

mikk150:
Emby has same issue as well

fulder:
After searching and digging around a bit after the cause for this issue I think I finally managed to find something. It look like mono on 12.1 version is a bit different build compared to e.g. 12.2 version. 

12.1:
```
Mono JIT compiler version 5.10.1.57 (5.10.1.57 Wed Feb 17 21:18:53 UTC 2021)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           __thread
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          supported, not enabled.
	GC:            sgen (concurrent by default)
```

12.2:
```
Mono JIT compiler version 5.10.1.57 (5.10.1.57 Tue Feb 16 01:44:31 UTC 2021)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           __thread
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          supported, not enabled.
[...]

PR #169, opened 2021-02-21
Cirrus CI tasks for plugin installation tests

fulder:
Add Cirrus CI tasks for testing installation of all plugins in this repository. 

The main reason and motivation for this PR and implementation is to get an overview of working and non-working plugins in this repository without being forced to test it manually or wait for issues from users (e.g. running a newer version of FreeBSD). This will also make it easier to know if updating a lot of plugins at the same time (for example when EoL of FreeBSD/plugin is reached and the release/artifacts updated) doesn't break anything for the end users during updates.

Except from installation feedback, these tests also give feedback about how long time it takes to install a plugin. This can be useful in order to optimize and minimize the installation progress (as an example: I have already noticed I installed `git` instead of `git-lite` in some of my plugins) 

I started out trying to implement this using GitHub actions but as there aren't any good VMs for FreeBSD (except for the MacOS VM one taking long time to boot up as well as only supporting 2 latest FreeBSD versions: https://github.com/marketplace/actions/freebsd-vm) I switched to Cirrus CI, it has support for a lot of different FreeBSD versions (out of the box): https://cirrus-ci.org/guide/FreeBSD/#list-of-available-image-families.

Furthermore, during the sh task implementation I realized it got more and more alike another "iocage plugin implementation" (although much more hardcoded than that). It would probably work OK to simply install the iocage from `pkg`, but as it does have a lot of own package dependencies it could make the tests less accurate (as forgetting a plugin dependency being available in iocage dependencies would be missed), as well as an `iocage` update/bug could make the tests fail incorrectly. That said testing `iocage` CLI at the same time doesn't feel fit for this repository and would probably be done better in the official `iocage` repo.

This PR will run cirrus task for all currently available plugins, this is done only during the initial/first implementation of this pipeline in order to get a first/initial build status for all the plugins and the master branch. Running all tasks all the time is a bad idea (especially if no plugins have been added/updated) to do for every following PR due to long run time and Free Cirrus public github repositories limitations: https://cirrus-ci.org/faq/#are-there-any-limits. A new PR will follow if/after this gets merged only running plugin tasks on changes.

If you decide to merge this PR (or enable the run from this PR), this repo will need to get the Cirrus CI github plugin installed. See official quickstart guide here for the (simple) required steps: https://cirrus-ci.org/guide/quick-start/

I've tried to install it on my own fork, see latest builds: https://cirrus-ci.com/github/fulder/iocage-plugin-index

There are currently a few plugins already failing in these tests because of different kind of errors (FreeBSD EoL, invalid packages/versions, no longer maintained plugins), I have another branch fixing the most common errors (which should probably solve some of the currently open issues in this repo). But IMO it should be OK to merge this although the check is failing and fix the errors in new PRs one at a time by/with support from plugin owners.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,8 +1,3 @@
-freebsd_instance:
-  image_family: freebsd-12-2
-  cpu: 1
-  memory: "1"
-
 install_template: &INSTALL_PLUGIN
   requirements_script: pkg install -y jq
   install_script: |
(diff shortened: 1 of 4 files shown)


tprelog:
Thanks for putting this together @fulder -- Hopefully we see this get approved.

---

> get an overview of working and non-working plugins in this repository without being forced to test it manually or wait for issues from users (e.g. running a newer version of FreeBSD).


I do not use every plugin I maintain, so this will be a huge help for me - Currently, I try to test them manually every few months. But it's easy to get lazy, side tracked or just forget.

---

> This will also make it easier to know if updating a lot of plugins at the same time (for example when EoL of FreeBSD/plugin is reached and the release/artifacts updated) doesn't break anything for the end users during updates.

A few of my plugins are not a typical FreeBSD packages that can just be installed or upgraded using `pkg`

Let me pick on my most popular plugin, (which by the way, I see has failed some step in you test)

I'm not familiar with any CI, so this is all like magic to me :mage: -- Can you say if your test is failing because it is trying to upgrade the plugin from 11.x --> 12.x -- For Home Assistant, this upgrade is not so simple because I gave a full refactor to the supporting scripts, and changed some defaults since the 11.3 - Basically, I guess it was major version change to the plugin itself.

Again, I have no idea what's involved in making this (it looks like a lot) so please forgive me if this is asking for to much...

Can (or will) this show separate results for a clean install versus upgrade?

I now recommend [using external dataset for the configuration](https://github.com/tprelog/iocage-homeassistant/wiki/external_dataset) - I'm thinking, with the config (of any plugin) moved to an external dataset, it's rather trivial to just reinstall a fresh copy of a plugin, rather than trouble shoot a failed update / upgrade attempt. Having test results split might give a better overview of a plugins situation.

Please consider (Let's pretend we have FreeNAS plugins for RELEASE-11.4)

**Example A**

- plugin installs using `RELEASE-11.4` :heavy_check_mark:
- plugin upgrades from `RELEASE-11.4` to `RELEASE-12.1` :x: 
- plugin installs using `RELEASE-12.1` :heavy_check_mark:

result - user know's a clean install using most recent very will work, even if the upgrade fails


**Example B**

- plugin installs using `RELEASE-11.4` :heavy_check_mark:
- plugin upgrades from `RELEASE-11.4` to `RELEASE-12.1` :x: 
- plugin installs using `RELEASE-12.1` :x: 

result - user know it is not possible to run an updated release. There is a more serious issue / plugin broke / something needs fixed / get someone's attention.

---

For the record - I have just tested a clean install locally, using both 12.1 and 12.2 for my plugin that have failed the CI

I have reached different results

**Home Assistant**

CI test
image

local test
image

- RELEASE 12.1
image

- RELEASE 12.2
image


**Gogs**

CL test
image

local test
image


- RELEASE 12.1
image

- RELEASE 12.2
image

tprelog:
:thinking: - Well that time seems a little... excessive

image

---

I found the issue with Gogs --  I think... the CI that's breaking this one :rofl:

image

Gogs will install `git` as a dependency, so neither `git` or `git-lite` is in the plugin manifest. Only `gogs` - I can add `git-lite` to the manifest though.

fulder:
Thank you very much for all the detailed feedback, questions, manual tests and good comments @tprelog.

I'm happy this PR could be helpful for you as well.

>I'm not familiar with any CI, so this is all like magic to me mage -- Can you say if your test is failing because it is trying to upgrade the plugin from 11.x --> 12.x -- For Home Assistant, this upgrade is not so simple because I gave a full refactor to the supporting scripts, and changed some defaults since the 11.3 - Basically, I guess it was major version change to the plugin itself.

Currently the tests are trying to mimic the real installation (probably with some new bugs I haven't noticed before your comment) in the following way/steps:
* Parsing the plugin JSON file
* Updating the `pkg` to use the specified `PACKAGESITE` together with the plugin fingerprints
* Installing all the plugin packages 
* Cloning the plugin repo (using `git-lite` which is installed before cloning and uninstalled after)
* Running the `post_install.sh` script
* Running the `pre_update.sh` and `post_update.sh` scripts if present (in order to see if these scripts work just after a clean update and using the same FreeBSD version)
* (I'm currently adding support to testing to fetch the admin portal specified in `ui.json` as there has been issues where the installation is "successful" but plugins don't start and can not be reached at all)

Hope this clarifies what is done in the CI behind the scenes a bit more :)

To comment on the versions a bit more, I was thinking it is up to the plugin creators to decide if they want to support older versions (and add them to the test matrix) or just the latest which should then be upgraded together with upgrading the JSON release.

> Again, I have no idea what's involved in making this (it looks like a lot) so please forgive me if this is asking for to much...

No problem, happy to explain how I've thought while creating these CI tests. Good to know if I haven't thought completely wrong about this or missed something.

>Can (or will) this show separate results for a clean install versus upgrade?

It should probably be possible to split this up more into an clean install and upgrade tasks (between different kind of version), but guess this is something we could do in future PRs/improvements. This is just a first try of a happy case installation and upgrade on the same FreeBSD version.

> I found the issue with Gogs -- I think... the CI that's breaking this one rofl

Good catch! The problem there is probably the plugin packages being installed before the "temporary" git for the plugin clone which then cleans up too many of the real plugin dependencies. I've switched the pkg installation order a bit in the latest commit (so git-lite/git is installed + the repo cloned and then cleaned up before the plugin packages). Let's see if this helps the actually properly working plugins.

.cirrus.yml
```diff
@@ -0,0 +1,565 @@
...
+  env:
+    PLUGIN_FILE: "heimdall-dashboard.json"
+
+homeassistant_task:
+  <<: *INSTALL_PLUGIN
+  matrix:
+    - freebsd_instance:
+        image_family: freebsd-11-2
+    - freebsd_instance:
+        image_family: freebsd-11-3
```
tprelog, line 254: I'm surprised this even passed the test on an 11.x 

However, as the plugin's maintainer, I am not long supporting the Home Assistant plugin on an 11.x release
```suggestion
```

.cirrus.yml
```diff
@@ -0,0 +1,565 @@
...
+    - freebsd_instance:
+        image_family: freebsd-12-1
```
tprelog, line 256: I'm surprised this even passed the test on an 11.x

However, as the plugin's maintainer, I am not long supporting the Home Assistant plugin on an 11.x release

```suggestion
    - freebsd_instance:
        image_family: freebsd-12.1
    - freebsd_instance:
        image_family: freebsd-12-2
```

fulder: Added this one as a debug step, because I got a bit confused why it was failing myself. Should be removed and match the plugin JSON file now.

fulder:
>I now recommend using external dataset for the configuration - I'm thinking, with the config (of any plugin) moved to an external dataset, it's rather trivial to just reinstall a fresh copy of a plugin, rather than trouble shoot a failed update / upgrade attempt. Having test results split might give a better overview of a plugins situation.

This is a nice idea! :+1: Haven't thought about this myself. Knowing if a plugin can be upgraded from one major FreeBSD version to another is tricky and hard to test (if e.g. a user has an unique plugin configuration which is the one breaking the upgrade).

fulder:
>:thinking: - Well that time seems a little... excessive

This does look like a bit too slow :open_mouth: . Can see in the logs that all the npm install stuff took around 4m so not really sure where all the extra 13 m. are coming from. Right now the FreeBSD VMs are using 1 vCPU and 1 GB Memory (which was the minimal resources I could set), it feels like it should be enough for a plugin installation, but if needed these limits could be increased. Although it would be a good idea to do it after the PR only running tests during the JSON plugin updates (otherwise the initial task run will be even more slow as more tasks will be queued up).

.cirrus.yml
```diff
@@ -0,0 +1,559 @@
...
+
+    if [ -f ./plugin/pre_update.sh ]
+    then
+      . ./plugin/pre_update.sh
+    fi
+
+    if [ -f ./plugin/post_update.sh ]
+    then
+      . ./plugin/post_update.sh
+    fi
```
tprelog, line 94: Just wondering why you decided to source these scripts instead of just running them.

I have another coner case for you - I have _a lot_ of corners, btw :stuck_out_tongue_closed_eyes:

I suppose you could say I cheat, err have some not so typical ways. That is to say, I use `sysrc`... a lot -- So, I have `. /etc/rc.subr && load_rc_config` included with some scripts that may be used during an install or update  -- basically some of my `post_*.sh` scripts are calling other scripts. Here, I have a separate install script because this same script is also used for the upgrades - having the script separate allow it to used during plugin install and dates, and also the provided console menu which allows the service to upgraded at any time, without running a plugin update

When viewing the command output....

- when the post scripts are run directly - [see output](https://cirrus-ci.com/task/5421363269730304?command=install#L162)

image




- when the post scripts are being sourced or `.` - [see output](https://cirrus-ci.com/task/6234190262829056)

image

You may notice, in the latter output, the entire `/etc/rc.subr` script is included


It all works the same - again this is a question, not a complaint

tprelog:
> I'm happy this PR could be helpful to you as well...
> Hope this clarifies what is done in the CI behind the scenes a bit more :)

Yep, regardless of what happens here, I'm going to use this in my own repos...



> Right now the FreeBSD VMs are using 1 vCPU and 1 GB Memory (which was the minimal resources I could set), it feels like it should be enough for a plugin installation, if needed, these limits could be increased.

This is more like it :joy:
```yaml
freebsd_instance:
  cpu: 4
  memory: "4"
```
image

.cirrus.yml
fulder: I was debugging why the homeassistent plugin still failed, but reverted to normal execution instead of just sourcing the files. Will debug some more on another branch, sorry for the confusion. Was thinking there was something wrong during the source of virtualenv inside the homassistant service install script. But realized it is run as another user (using `su` ) entierly so shouldn't matter.

A bit confused why the venv can not be found during the installation right now.

fulder:
> This is more like it :joy:

Nice! Then looks like more resources does solve the problem after all. These values can probably be boosted up pretty much if this goes live and not all plugins are executed every run :crossed_fingers:

tprelog:
> Nice! Then looks like more resources does solve the problem after all. These values can probably be boosted up pretty much if this goes live and not all plugins are executed every run

Just for a  comparison

- I'm not going to try and math out the percentages here - but doubling the initial resources did not cut the time in half.
```
freebsd_instance:
  cpu: 2
  memory: "2"
```

image

.cirrus.yml
tprelog: > why the homeassistent plugin still failed

`usr/local/etc/rc.d/homeassistant` - Perfect example of my not so typical way's, lol

tprelog:
So I'm guessing there maybe something going on in the 12.1 release -- but it's working on 12.2 :+1: 

**Interesting:**

image

image

image

fulder:
> So I'm guessing there maybe something going on in the 12.1 release -- but it's working on 12.2 :+1:

Godd findings! :+1: `esphome` seems to have the exact same issue as `homeassistant`, I think there are some issues with the venv after all. Printing some more information (after post_install script) I got the following.

```
./plugin/post_install.sh || true
plugin_initialized:  -> 5.0.20210222
plugin_version:  -> 5.0.20210222
homeassistant_umask:  -> 002
homeassistant_user:  -> homeassistant
homeassistant_group:  -> homeassistant
homeassistant_config_dir:  -> /home/homeassistant/homeassistant
homeassistant_python:  -> /usr/local/bin/python3.8
homeassistant_venv:  -> /usr/local/share/homeassistant
Error: Command '['/usr/local/share/homeassistant/bin/python3.8', '-Im', 'ensurepip', '--upgrade', '--default-pip']' returned non-zero exit status 1.
_: line 4: /usr/local/share/homeassistant/bin/activate: No such file or directory
/usr/local/etc/rc.d/homeassistant: ERROR: install function failed

whoami
root
ls -l /usr/local/share/homeassistant
total 32
drwxrwxr-x  2 homeassistant  homeassistant  512 Feb 22 00:36 bin
drwxrwxr-x  2 homeassistant  homeassistant  512 Feb 22 00:36 include
drwxrwxr-x  3 homeassistant  homeassistant  512 Feb 22 00:36 lib
lrwxrwxr-x  1 homeassistant  homeassistant    3 Feb 22 00:36 lib64 -> lib
-rw-rw-r--  1 homeassistant  homeassistant   75 Feb 22 00:36 pyvenv.cfg
ls -l /usr/local/share/homeassistant/bin
total 0
lrwxrwxr-x  1 homeassistant  homeassistant   9 Feb 22 00:36 python -> python3.8
lrwxrwxr-x  1 homeassistant  homeassistant   9 Feb 22 00:36 python3 -> python3.8
lrwxrwxr-x  1 homeassistant  homeassistant  24 Feb 22 00:36 python3.8 -> /usr/local/bin/python3.8
service homeassistant oneinstall
/usr/local/etc/rc.d/homeassistant: ERROR: /usr/local/share/homeassistant/bin/activate is not found
```

So it is (partly?) installed but missing the `activate` binary. 

Although trying to install the plugin manually on my TrueNAS and using 12.1 release work just fine, which is weird. The releases should be the same version, using the same packagesite etc. I've even tried to just boot up a totaly empty jail, install only python, bash and git-lite, cloning the repo and running the `post_install` manually. Still worked without any problems.

.cirrus.yml
```diff
@@ -0,0 +1,559 @@
...
+
+    if [ -d "./plugin/overlay" ]
+    then
+      echo "Found overlay folder"
+      cp -r ./plugin/overlay/ /
+    fi
+
+    ./plugin/post_install.sh
+
+    if [ -f ./plugin/pre_update.sh ]
```
tprelog, line 86: _I think_ the `post*.sh` scripts need to be executable, Maybe we should check for that here
```suggestion
    if [ -x ./plugin/pre_update.sh ]
```

.cirrus.yml
```diff
@@ -0,0 +1,559 @@
...
+    then
+      ./plugin/pre_update.sh
+    fi
+
+    if [ -f ./plugin/post_update.sh ]
```
tprelog, line 91:
```suggestion
    if [ -x ./plugin/post_update.sh ]
```

sonicaj (commented):
Thank you for your contribution, good work :)

We have `kmods` as well which can be specified in a manifest and we are not adding them right now. Please see https://github.com/truenas/iocage/blob/bfb1db70da989ae20b580f5fac04389d0905c2f5/iocage_lib/ioc_plugin.py#L611
for how iocage does it. Also i wonder if using `https` packagesite will work currently or if `ca_root_nss` is installed by default.

.cirrus.yml
```diff
@@ -0,0 +1,559 @@
...
+    then
+      echo "Found overlay folder"
+      cp -r ./plugin/overlay/ /
+    fi
+
+    ./plugin/post_install.sh
+
+    if [ -f ./plugin/pre_update.sh ]
+    then
+      ./plugin/pre_update.sh
```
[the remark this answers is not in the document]
tprelog: I found another corner case in the home assistant plugin, that shows a [false negative](https://github.com/tprelog/test/runs/1957637076) --- If you look before the error, you can see the service has started and status shows it's running -- ( what I'm doing there is OT, but I have another solution for that in the zwavejs2mqtt plugin )

On the other hand - maybe this is not really _false_ negative because it shows a user can not expect to run the upgrade after the fact -- I mean, is it reasonable to think if a plugin can not run the update scripts on a clean install, it is not going to work when the user tries to update the plugin after later on?

Even if we do not run the scripts, I'm fine with that -- Should we still check if they are at least executable when they do exists?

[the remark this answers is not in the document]
sonicaj: Yes we should / can still check if they are executable :)

tprelog:
I'm in the waters of SCALE, heading out to sea in search of new lands.... Just as the shore is about to leave sight, something catches my eye... It is the plugins on CORE, waving their hands and screaming WAIT... COME BACK... _We have a cirrus CI to automate testing now !!_ :joy: :joy: 

Seriously though, this changes everything for me....

.cirrus.yml
fulder: I was looking a lot through this file to understand what was really happening in there :smile: . Haven't had much time to debug homeassitant much more but my latest theory is that the venv module has different version in FreeBSD 12.2 (and 12.2-p2) compared to 12.1 hence the different result.

.cirrus.yml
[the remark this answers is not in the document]
fulder: During the first implementation of these tests I was thinking it would be a good idea (like @tprelog commented) to test if the users would be able to upgrade the plugins just after a clean installation (that's the reason behind these executions). On the other hand, it is maybe not a probable test case after all, I'm guessing the majority of users upgrade a plugin mostly because of a FreeBSD version upgrade.

Have no problem removing this for now and only check if the scripts are executable (I know I've personally forgotten to change the mode of these, and the post_install, scripts in my plugins), should be changed now.

.cirrus.yml
fulder: Should be updated to `if ! [-x ./plugin/pre_update.sh ]` with and `exit 1` if that's true. Do you think that's fine?

.cirrus.yml
fulder: Should be updated to `if ! [-x ./plugin/post_update.sh ]` with and `exit 1` if that's true. Do you think that's fine?

fulder:
>We have kmods as well which can be specified in a manifest and we are not adding them right now. Please see https://github.com/truenas/iocage/blob/bfb1db70da989ae20b580f5fac04389d0905c2f5/iocage_lib/ioc_plugin.py#L611
for how iocage does it. Also i wonder if using https packagesite will work currently or if ca_root_nss is installed by default.

Thank you for your review and good comments @sonicaj! :handshake: 

Interesting about `kmods` and `https` protocol for packagesite. Didn't know `iocage` had support for that (sorry for not being really familiar with the `iocage` project, I've mostly followed the current plugin manifests in order to figure out what the different JSON properties were used for, could maybe be cool with some documentation about all the supported required/optional properties :angel:). 

I've tried to add support for `kmods` (https://github.com/ix-plugin-hub/iocage-plugin-index/pull/169/commits/b3b05c4c4108fb68c3a40cd54a4c5bb0ce3b757f) and installing (+ removing before plugin pkg installation) the CA root certificates (https://github.com/ix-plugin-hub/iocage-plugin-index/pull/169/commits/9e8976a522d79222327e3c35b00a000da490871e) shouldn't hurt the current plugin tests. Will have to test these 2 things with example plugins more properly though.

.cirrus.yml
fulder: https://cirrus-ci.com/task/6055689542959104 tried to run the failing (ensurepip) command manually in 12.1. Looks like it is the "common" Symbol variable issue. Probably related to e.g. this issue:  https://github.com/ix-plugin-hub/iocage-plugin-index/issues/165. 

(This issue was actually another motivation for this PR, I was thinking about bumping a lot of 12.1 plugins into 12.2 but then I got scared thinking "what if I break something" :stuck_out_tongue:)

Ornias1993:
Just came across this... Nice work on that CI @fulder I can appreciate good CI work! :)

One thing that should also be noted:
Maybe it's also time to document when plugins are too unmaintained and broken to still be included?
I've come across a few that are like YEARS without any TLC...

fulder:
Thank you @Ornias1993 :+1: 

That's a good point indeed, although can be a bit hard to decide what to do with that kind of plugins. Maybe they should simply be removed and clean up from this index. Having a non-working plugin with no issues about it nor any actions (during a long period of time) can be more misleading than not having it in the index even if it was installed some long time ago.

fulder:
I've added some refactoring to the `pkg` commands, hopefully making it easier to understand and fixing potential bugs/problems. (Sorry for many commits after your already started/done reviews but looks like more changes than it really is :angel:)

The `pkg` flow after this is now:

1. Install `jq` needed for manifest parsing
2. Install `git-lite` (or if not present fall back to `git`) in order to clone the repo artifacts
3. Delete ALL installed packages, run `autoremove` and `clean` to start out without anything (except pkg itself) installed before plugin `pkgs` installation
4. Install plugin packages:
    1. Install `ca_root_nss` if packagesite contains `https`
    2. Switch to plugin packagesite and fingerprints
    3. Run `pkg fetch`
    4. Remove the previously installed `ca_root_nss` 
    5. Install all the fetched and cached packages

-----

**Potential bug and fixes:**

Previously all the packages (installed on cirrus, actually GCP, VM) were not cleaned up before plugin pkg installations, this could mean forgetting a package included in these automatically installed packages would still make a plugin installation work, although a real installation would fail. 

The git CLI was installed AFTER the plugin packagesite/fingerprints were used. Using a packagesite where both are missing (or git-lite is missing) could make an installation fail. 

As only the installed test packages were installed (with some old if conditions) it would be easier to forget to clean something up as installing a new dependency would mean the list for `pkg delete` command would need to be updated as well.

.cirrus.yml
fulder: Had to check if it both exist and is executable because only checking with `-x` would result in failure if this optional script is missing. See error run e.g. here: https://cirrus-ci.com/task/5077739649630208.

Seems to work fine checking both conditions (as I wrote in the previous comment) see test run here: https://cirrus-ci.com/task/5443860110245888, so resolving this (as well as the review comment below), feel free to open it up again or add a new comment if this isn't correct.

.cirrus.yml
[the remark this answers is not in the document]
fulder: As these scripts are no longer executed I will resolve this. Feel free to open up or add a new comment if something is still unclear/feels wrong.

sonicaj:
Thank you @fulder for clarifying the new changes / current status. I wanted to confirm if the PR is ready for review now ?

fulder:
Yes @sonicaj , should be ready for review now again :) Latest run (commit: https://github.com/ix-plugin-hub/iocage-plugin-index/pull/169/commits/91e03f44297336668a029a31f3c0aeb11800c765) here: https://cirrus-ci.com/build/6584473031540736

sonicaj (approved):
Thank you, it looks in good shape to me :)

tprelog:
Thanks again for your work on this @fulder

tprelog:
So I found something (I think) It looks like TrueNAS is not matching correctly matching the release in the manifest to plugin's repository --- For example, the current `homeassistant.json` is set to `12.1-RELEASE` - so I push a 12.1-RELEASE branch to my plugin repo so I could start working on an update... I thought I should be able to work on updates using the master branch, and push to 12.2-RELEASE when ready, finished off with a pull request to have the release switch in the manifest (I have changes that need to be timed correctly with the manifest being updated)

anyway - I spent days trying to fix and final test against things 12.1-RELEASE only to realize that the plugin is trying to install from the 12.2-RELEASE even thought the manifest says 12.1-RELEASE


```
$ sudo iocage fetch -P homeassistant.json --name ha-test-default
Plugin: homeassistant
  Official Plugin: false
  Using RELEASE: 12.1-RELEASE
  Using Branch: 12.2-RELEASE
  Post-install Artifact: https://github.com/tprelog/iocage-homeassistant.git
  These pkgs will be installed:
    - autoconf
    - bash
```

Currently this results in the plugin installing from master --  I was about to push changes to my master branch -- 

I would have broke everything... :disappointed_relieved: 

I bring this up here because _this CI is using the correct branch_ to install from... 

I just want to mention this is something to watch out for,  if were expecting the artifact repo to match the release set in the manifest

tprelog:
so, looks like TrueNAS is not using the release set in the manifest, but instead, it's using the branch the manifest have been git from

```
$ git branch 
* 12.2-RELEASE
  master
```

( that's my `iocage/.plugins/github_com_ix-plugin-hub_iocage-plugin-index_git` directory BTW)

Is there any way to issue ( even in a future pull request ) that we can issue a warning when this branch does not match the release set in the manifest - I  don't think it should fail a build because of this, but it could be useful to throw an alert, since there is no ( at least not that I'm aware of ) notice given to plugin creators when these branches get switched

fulder:
Oh good (and scary :fearful: ) finding @tprelog. Also a good example where an `iocage` bug could block plugin development/merge if we used it in this CI. 

Would still be a good idea to test plugin installation using `iocage` CLI instead, but as written in this PR description, this could maybe be done in the (already present) cirrus CI in the `iocage` repository (e.g. using a test/dummy plugin). That test could then verify if things like you've just done manually work as intended :)

I've tried to install `homeassistant` plugin on my private TrueNAS machine and verified this bug:

```
Plugin: homeassistant
  Official Plugin: false
  Using RELEASE: 12.1-RELEASE
  Using Branch: 12.2-RELEASE
```
and after `pkgs` installation:

```
Branch 12.2-RELEASE does not exist at https://github.com/tprelog/iocage-homeassistant.git!
Using "master" branch for plugin, this may not work with your RELEASE
```

After digging around in the `iocage` code I hope I found the issue. Could maybe this be a solution for the bug: https://github.com/iocage/iocage/pull/1248? 

After this change and re-running the `iocage fetch` command I didn't get the clone warning anymore:

```
Fetching artifact... 
Cloning git repository

Running post_install.sh
```

And checking the plugin fetch location I could see it was now at `12.1-RELEASE` branch.

fulder:
>I bring this up here because this CI is using the correct branch to install from...

That's a good point, it does differ in the matter of cloned version. In this repository and tests we cannot know what TN/FN machine this plugin will be installed on. Maybe we should add an extra and optional parameter to the tests in order to override the plugin repo branch (otherwise still defaulting to the VM version) and to make them more real today's installation. 

Even though the `iocage` PR updating the plugin artifact branch to be equal plugin version would be merged, having the flexibility to change the plugin artifact branch in the tests can be useful (e.g. in order to check if all the release branches really are needed, or during developmen using a non-standard branch). I can start to work on that but on another branch (and we can maybe move to the `iocage` PR for further discussion how the artifact branch should be handled), this one has already been approved and has started to grow a lot (with new branches based on it) so maybe it's time to merge this one now? I'm sure we can iterate over the not working plugins and improve the tests (in any possible way) after this initial PR.

As a reminder (to whoever merges this, @kmoore134 ?) would be cool if you could enable the [cirrus-ci github plugin](https://cirrus-ci.org/guide/quick-start/) for this repository before merge so we can start getting some task runs on PRs/master :)

tprelog:
@fulder on last thing... might as well update my plugins to use the 12.2-RELEASE in the CI test. I see the 12.1-RELEASE is EOL now anyways - I'll have a separate pull request later today, changing the RELEASE in the corresponding plugin manifest.

I'll make the suggestion above

.cirrus.yml
```diff
@@ -0,0 +1,579 @@
...
+    - freebsd_instance:
+        image_family: freebsd-12-2
+  env:
+    PLUGIN_FILE: "emby-server-stable.json"
+
+esphome_task:
+  <<: *INSTALL_PLUGIN
+  matrix:
+    - freebsd_instance:
+        image_family: freebsd-12-1
```
tprelog, line 216:
```suggestion
        image_family: freebsd-12-2
```

.cirrus.yml
```diff
@@ -0,0 +1,579 @@
...
+  env:
+    PLUGIN_FILE: "gitlab.json"
+
+gogs_task:
+  <<: *INSTALL_PLUGIN
+  matrix:
+    - freebsd_instance:
+        image_family: freebsd-12-1
```
tprelog, line 248:
```suggestion
        image_family: freebsd-12-2
```

.cirrus.yml
```diff
@@ -0,0 +1,579 @@
...
+  env:
+    PLUGIN_FILE: "heimdall-dashboard.json"
+
+homeassistant_task:
+  <<: *INSTALL_PLUGIN
+  matrix:
+    - freebsd_instance:
+        image_family: freebsd-12-1
```
tprelog, line 272:
```suggestion
        image_family: freebsd-12-2
```

.cirrus.yml
```diff
@@ -0,0 +1,579 @@
...
+    - freebsd_instance:
+        image_family: freebsd-11-3
+  env:
+    PLUGIN_FILE: "monica.json"
+
+mosquitto_task:
+  <<: *INSTALL_PLUGIN
+  matrix:
+    - freebsd_instance:
+        image_family: freebsd-12-1
```
tprelog, line 368:
```suggestion
        image_family: freebsd-12-2
```

.cirrus.yml
```diff
@@ -0,0 +1,579 @@
...
+  env:
+    PLUGIN_FILE: "netdata.json"
+
+node-red_task:
+  <<: *INSTALL_PLUGIN
+  matrix:
+    - freebsd_instance:
+        image_family: freebsd-12-1
```
tprelog, line 392:
```suggestion
        image_family: freebsd-12-2
```

.cirrus.yml
fulder: Sure thing, I was making this change as well on one of my test branches.

.cirrus.yml
```diff
@@ -0,0 +1,579 @@
...
+  env:
+    PLUGIN_FILE: "sonarr.json"
+
+tasmoadmin_task:
+  <<: *INSTALL_PLUGIN
+  matrix:
+    - freebsd_instance:
+        image_family: freebsd-12-1
```
tprelog, line 496:
```suggestion
        image_family: freebsd-12-2
```

tprelog:
OK last one, for real this time... I like the shields, lol

How about this at top of the README (both are static) It's a little eye candy and provide in simple way to communicate which branch is default. When the branch changes, just change the label in the shield... It's a simple solution for now

[plugins-link]:https://www.truenas.com/plugins/
[plugins-shield]:https://img.shields.io/badge/TrueNAS%20CORE-Community%20Plugin-blue?logo=TrueNAS&style=for-the-badge
[release-shield]:https://img.shields.io/badge/Default%20Branch-12.2--RELEASE-blue?logo=FreeBSD&logoColor=red&style=for-the-badge

[![x][plugins-shield]][plugins-link] [![x][release-shield]][release-shield]

```
[plugins-link]:https://www.truenas.com/plugins/
[plugins-shield]:https://img.shields.io/badge/TrueNAS%20CORE-Community%20Plugin-blue?logo=TrueNAS&style=for-the-badge
[release-shield]:https://img.shields.io/badge/Default%20Branch-12.2--RELEASE-blue?logo=FreeBSD&logoColor=red&style=for-the-badge

[![x][plugins-shield]][plugins-link] [![x][release-shield]][release-shield]
```

fulder:
>OK last one, for real this time... I like the shields, lol

These were purty indeed :smile: pushed them in as an extra thing to this PR. Also added a link to the release shield to link to the FreeBSD release page instead of the shield iteself.

sonicaj:
@fulder just to be sure, is everything green from your end wrt this PR ? We can have this in then :)

fulder:
@sonicaj yes it should be green and ready :) will probably want to merge https://github.com/ix-plugin-hub/iocage-plugin-index/pull/170 after this one, but it doesn't contain as many changes at all compared to this PR. Feel free to merge when you feel ready =)

PR #170, opened 2021-02-21
Run Cirrus CI tasks on changes only

fulder:
**Note**: Review/Merge after: https://github.com/ix-plugin-hub/iocage-plugin-index/pull/169

Run tests only if the plugin `.json` file is changed

**EDIT**:

Changes include:
* Bumping the VM machines from 1 CPU and 1 GB RAM to 4 CPUs and 4 GB of RAM. After this change a plugin test will be run only if the plugin manifest is changed and probably 1 to a few plugins are changed in the majority of cases in this repository PRs. This is also done becuase we previously saw issues with slow running tests which can give false impression a plugin installation takes too long time.
* Move the install script into own file and trigger all tests on changes. This file shouldn't change much when the test gets more and more stable. Still having the script inside the `.cirrus.yml` file and trigger all plugins on these changes would make one plugin task update re-run all plugin tests which isn't really the expected behavior. This will also give better syntax highlighting for shell instead of having it inside a YAML file. 
* Run the install script with `set -e` in order to fail the execution if something goes wrong in there
* Install `jq` as part of the script instead of another setup step.


sonicaj:
@fulder please do let me know once this PR is ready for review. Thank you

fulder:
@sonicaj thank you for merging the previous large PR into master :+1: 

Was expecting this PR commits to decrease after that was done but realized you squashed all the commits during merge. That's just fine for my part (you can squash them from this branch as well when it's time). I've just merged upstream master into this fork and updated the description with changes done. It should be ready for review/merge now.

sonicaj (approved):
LGTM, thank you for your contribution. It's really appreciated :)

fulder:
Looks like the test-run results have 1 more success compared to current `master`. `clamav` now passed, but it failed before, the test results doesn't seem much different. Only seems to have run very much faster here. 

Think it should be fine to merge this one in as well.

PR #175, opened 2021-03-01
Fix plugin test failures

fulder:
Fixes some of the current master test issues:

* `calibre-web` is having issues with python being compiled for the newest FreeBSD version (`close_range@FBSD_1.6`). Upgraded it to 12.2 in order to solve the issue but the tests are currently still failing due to the `pre_update` script being not executable. I've created a new PR in the plugin repo for this issue: https://github.com/aserrallerios/iocage-plugin-calibre-web/pull/2. Also missing needed `rust` package in order to be able to compile the python `cryptography`.
* `deluge-pip` has the same `close_range@FBSD_1.6` issue, updating it to 12.2 made the tests for this plugin green. Also missing `rust` just like `calibre-web`.
* `monica` packages seems outdated, so have updated them to the latest minor version
* `motioneye` had an old and no longer existing python version
* `tautulli` `close_range@FBSD_1.6` => upgraded to 12.2
* `unificontroller` seems to have a too recent package version, have downgraded it

After these fixes there are still 3 plugins with failures:
* `emby` fix in: https://github.com/ix-plugin-hub/iocage-plugin-index/pull/176
* `calibre-web` see text above
* `openvpn` gets stuck during post_install script for some reason (probably better the plugin maintainer look into this one)

---

An interesting test bug (which has been discussed a bit in the first PR) to point out is the `motioneye` where the current changes fixes one issue with the package installation but because there are other problems related to the `post_install.sh` scripts (these are not caught because it is ignoring all the errors) this gives a false positive result of the test run. In order to catch these kind of bugs I will have a PR ready shortly with Admin UI check available (this has also been vaguely discussed before). The UI branch is based on https://github.com/ix-plugin-hub/iocage-plugin-index/pull/170 so would be nice to merge that one soon :)


sonicaj (approved):
Thank you @fulder for your contribution :)

jsegaert:
@fulder Some fallout... MotionEye has some hardcoded references to pip-2.7 in its post_install script.
https://www.truenas.com/community/threads/community-plugins-motioneye.91652/

fulder:
@jsegaert thanks for poitning that out. See related PR and discussion here: https://github.com/ix-plugin-hub/iocage-plugin-index/pull/188#issuecomment-797423576

PR #177, opened 2021-03-02
Fix consistent README badges

fulder:
README updates:
* Use same shield style for all badges
* Remove plugin specific badges. These return 504 codes sometimes there's a good summarization about plugin runs both at the master commit check and the cirrus ci github report e.g. here: https://github.com/ix-plugin-hub/iocage-plugin-index/runs/2002402831

This way new plugins won't have to update this long README list as one of the steps.


kmoore134 (approved):

kmoore134:
This looks much nicer, thanks!!

Issue #138, opened 2020-12-20
rtorrent-flood webui does not start


kimothy:
rtorrent-flood webui does not load when trying to access it through the web browser. It worked in Freenas 11.2, however not in Truenas Core.

greysave:
I have the same issue.

greysave:
> rtorrent-flood webui does not load when trying to access it through the web browser. It worked in Freenas 11.2, however not in Truenas Core.

I have the same issue.  I need this to work.  My work around right now is to use qbittorrent, with openvpn manually installed.  This is a better solution if possible.

fulder:
Related to issue: https://github.com/nodejs/node-gyp/issues/2048 the plugin is using an old version of `node-gyp` which doesn't have support for python 3.X. I've created a pull request in the plugin repository for a possible fix.

fulder:
I've made an attempt of upgrading the flood repo from the old abandoned one to the new one in: https://github.com/pierrickrouxel/iocage-plugin-rtorrent-flood/pull/8. Managed to get rtorrent + flood up and running using this branch (and an updated manifest, a PR will follow in this repo if it gets approved)

fulder:
The PR is now merged and installation test (see e.g.: https://cirrus-ci.com/task/5610691806625792) is looking green again. @kimothy and @greysave if you could test if it works for you as well it would be much appreciated :)

PR #178, opened 2021-03-02
UI portal install test support

fulder:
Add UI portal support to plugin installation tests. This is mainly done to fix false positive tests (discussed in some of the old PRs). 

Summarized changes:
* If the plugin artifact contains the `ui.json` file with an URL (there are some plugins with just a informative text), replace the `%%IP%%` placeholder with the correct IP of the running VM (just using localhost/127.0.0.1 has been problematic with some of the plugins)
* If the `ui.json` contains additional `adminportal_placeholder` key/values inject the `settings.json` default values for them
* If `post_install.sh` script runs without any failure try to fetch the admin ui using `curl` (it has more helpful flags compared to `fetch` for this use case IMO)
* Plugins can add a `FOLLOW_REDIRECTS: false` in order to skip the auto redirecting on 3XX codes (and make the test success on this 3XX code). Jackett plugin is a good example where the UI is only open for external requests. Making an interna requests makes it loop in redirects forever.
* Plugins can add a `ADMIN_UI_USER` env variable to the tests if there is a curl user parameter needed in order to reach the UI (`nzbget` is a good example)

Extra changes:
* Color prints for success (green) / error (red) and info (purple) for easier debugging inside e.g. cirrus ui. Example run: https://cirrus-ci.com/task/4544229549539328

If you decide to merge this a new PR fixing some of the new failures will follow. 

Related issues:
* A bit related to issue: https://github.com/ix-plugin-hub/iocage-plugin-index/issues/165 as the new tests should be able to catch such issues (with unreachable admin portals)
* https://github.com/ix-plugin-hub/iocage-plugin-index/issues/138 this error is now making the `rtorrent-flood` crash with the same error as stated in this issue (this is currently false positive on master)


fulder:
One thing (a bit outside of this PR, but wanted just share my findings here) I've noticed during further debugging of tests and plugin installations is the FreeBSD kernel version. 

While installing a plugin on a real TN system the kernel version for TN will be used (even though the jail is actually running another FreeBSD version). This is probably the reason why we saw differences in e.g. the `homassistant` plugin runs @tprelog using the same `12.1-RELEASE` and when there were issues with the FreeBSD Symbol Variables only in the cirrus run. 

My guess is that because TN used the correct kernel version this symbol variables where correct in this case, but as the cirrus run had the old one it failed.

tprelog:
Nice job again @fulder

Home Assistant can take some extra time before the UI is reachable - I was worried it would not pass this test, but your implementation handles it!

> This is probably the reason why we saw differences in e.g. the `homassistant` plugin runs @tprelog using the same `12.1-RELEASE` and when there were issues with the FreeBSD Symbol Variables only in the cirrus run.

I agree, but I also think there is more to it... For example, even though Home Assistant installs (and starts) using the 12.1-RELEASE on TN, there are main integrations that are failing because some pip packages are failing to compile -- I also noticed in a `sed` command where I'm using `\n` but it does not correctly add a new line. (there is no error, it just adds an `n`)

Both issue are resolved after I changed to the 12.2-RELEASE in the manifest.

fulder:
Thank you @tprelog 

>Home Assistant can take some extra time before the UI is reachable - I was worried it would not pass this test, but your implementation handles it!

That's great to hear! :D I was thinking about just this problem while implementing this (have seen some plugins with admin UI not being accessible until a few seconds after the plugin installation completes). It does look pretty close with 5 retries and 5 s. timeout for `homeassistant` though: https://cirrus-ci.com/task/5251947076780032. Should maybe increase it a bit in order to not risk false failures.

> I agree, but I also think there is more to it... 
You're right that's probably more issues than I (and we) have seen/though about here. Good to just update it to the latest (and not EoL) FreeBSD version being in sync with the latest TN version.

fulder:
Feel free to review/merge this when you feel ready. I'm not doing anyting more on this branch/PR :)

sonicaj (approved):
Thank you for your contribution :)

fulder:
@kmoore134 can we merge this one? Should be ready. More UI fixes are available and ready in the (currently WIP marked) https://github.com/ix-plugin-hub/iocage-plugin-index/pull/179 based on this branch.

sonicaj:
@fulder can you please resolve conflicts and we can have this merged then ? Thank you

sonicaj:
@fulder please do let me know when you think this is safe to merge ( the conflicts have been resolved, but just want to be sure i have the go ahead from your side ;) )

fulder:
@sonicaj Merged upstream master and resolved the conflicts (with myself :sweat_smile:)

Double checked that both a service not started false positive test is still failing: https://cirrus-ci.com/task/6275917816266752?command=install#L353 as well as UI checks are still working after the conflict resolve: https://cirrus-ci.com/task/6425451397644288?command=install#L672. Don't have to wait for all the tests before merging IMO (a few will currently fail anyway). Feel free to merge this in now :bow:

Issue #118, opened 2020-10-13
Gitlab fails to install - FreeNAS-11.3-U5 - Release channel: 11.3-RELEASE


jdrames:
Since upgrading to FreeNAS-11.3-U5 I have been unable to install Gitlab from the Plugins page.

It appears that the install starts fine. However, on completion the output shows the following log messages:
```
Starting redis.
sed: /usr/local/www/gitlab-ce/config/gitlab.yml: No such file or directory
Starting redis.
sed: /usr/local/gitaly/config.toml: No such file or directory
Starting redis.
sed: /usr/local/www/gitlab-ce/config/database.yml: No such file or directory
Starting redis.
chown: /usr/local/www/gitlab-ce: No such file or directory
Starting redis.
cd: /usr/local/www/gitlab-ce: No such file or directory
Starting redis.
cd: /usr/local/www/gitlab-ce: No such file or directory
Starting redis.
cd: /usr/local/www/gitlab-ce: No such file or directory
Starting redis.
cd: /usr/local/www/gitlab-ce: No such file or directory
Starting redis.
cd: /usr/local/www/gitlab-ce: No such file or directory
Starting nginx...
cd: /usr/local/www/gitlab-ce: No such file or directory
Performing sanity check on nginx configuration:
cd: /usr/local/www/gitlab-ce: No such file or directory
```

I can verify that the /usr/local/www/gitlab-ce directory is missing from the jail. 
I did another install after this and while the install process was going on I monitored the www folder in the jail via shell and interestingly at some point the gitlab-ce directory is created but then later in the process it appears it is removed.

theozaurus:
I've run into this as well. I think it may be because package postgres11-server is specified and gitlab-ce. However, the latest gitlab-ce package needs postgres12-server.

jdrames:
Yeah, currently I've moved over to Gitea for the gitrepository and jenkins to the integration development/production pipeline. So I've got a working system currently in place. Perhaps once the FreeNAS(TrueNAS Core) move to ver 12 it'll be fixed. But with the current system I've got setup I might just skip on converting back to Gitlab.

theozaurus:
[...]

PR #179, opened 2021-03-03
UI portal fixes

fulder:
**Note**: review after: https://github.com/ix-plugin-hub/iocage-plugin-index/pull/178

TLDR:

This PR bumps some of the currently failing plugin to 12.2 to solve issues with `FBSD_1.6` symbol variable and fixes some small (both general and UI specific) issues with the installation tests.

It has also been the base for plugin repo PRs and fixes (see long progress/fixes list below for more info).

-----
Test fixes:
* Add `--retry-all-errors` to `curl` command. `transmission` plugin is a good example with some `Connection reset by peer` issues, but working after a few some more waiting time/retry. See test without the flag: https://cirrus-ci.com/task/5682271261818880 compared to with: https://cirrus-ci.com/task/5650314154999808
* Add support for skipping admin UI check in plugin tests using new env var: `SKIP_UI_CHECK: "false"` A good example is the `quasselcore` plugin containing admin portal URL (for the server IP/port) although it is no real UI and can only be reached using a quassel-client.
* Set global  `HOME` env var to `/root`. Starting an empty FreeBSD jail has this variable and this has been problematic for `homebridge` plugin where this variable is used for the `pm2` service.
* Set global  `USER` env var to `root`. Starting an empty FreeBSD jail has this variable and this has been problematic for `homebridge`/`hoobs` plugins where this variable is used for the `pm2` service.
* Disable `ipfw` after `post_install` script, if it has some rules disabling the communication between cirrus and the machine the job will just run forever until timeout (`openvpn` is a good example of this)
* Set global `PATH` env var for all possible binary locations (taken from the `iocage` repo). `openvpn` plugin does install the binary file inside the `/usr/local/sbin` which was previously missing from path.
* Improve wait for service function waiting a while for services to return proper status instead of just doing it once.
* Stop `ipfw` service after the service check. Plugins starting up this service in post install would otherwise always fail (as it would be stopped) and ignore the error logs when this service is not enabled.

Plugins upgraded to 12.2 due to `FBSD_1.6` symbol variable issues:
* `bazarr`
* `gitlab`
* `hoobs`
* `lidarr` 
* `radarr`
* `sonarr`
* `sickchill` (althought the test did pass, the symbol issue was still there: https://cirrus-ci.com/task/4812142425669632?command=install#L998)
* `homebridge`

Other fixes:
* `emby` admin UI not reachable even while installing the plugin manually (but can't see any error logs during installation). Upgrading it to 12.2 seems to have solved the issue.

Still existing bugs (left for plugin maintainers and/or other PRs), double checked if these plugins really don't work manually installing the plugins on my TN machine:
* Failing but with possible solutions (in plugin repos)
  * `calibre-web` UI not reachable. Created a possible plugin PR fix for this: https://github.com/aserrallerios/iocage-plugin-calibre-web/pull/3 
  * `jenkins` and `jenkins-lts` seems to have 5XX issues trying to access the UI. Possible fix in plugin PR: https://github.com/freenas/iocage-plugin-jenkins/pull/3
  * `motioneye` post_installation bugs. A bit different behavior here compared to manual installation (due to post_install bug the manual installation result in the plugin jail being destroyed again, but the test doesn't seem to crash because of this script, returning a 0 code even though there are some errors, although the UI is, as expected, not reachable after the installation). **Update**: Related PR/issues and possible solution (in PR comments): https://github.com/ix-plugin-hub/iocage-plugin-index/pull/188
  * ~~`rtorrent-flood` test still failing. Possible fix: https://github.com/pierrickrouxel/iocage-plugin-rtorrent-flood/pull/8.~~ **Update**: PR merged and seems like it's working again with the test on this branch being green.
  * ~~`gitlab` master run was green but had the same (missed) issues related to post_install.~~ **Update**: PR merged and seems like it's working again with the test on this branch being green.
* Failing with no solution right now:
  * ~~`homebridge` the plugin repository has been archived (https://github.com/grimneko/iocage-plugin-homebridge), maybe it's time to remove the plugin from this index? It did have the same 12.2 symbol issue but even after upgrading the UI can not be reached: https://cirrus-ci.com/task/6690476246958080~~
  * ~~`openvpn` there is something really buggy with the installation making it getting stuck spamming `Using SSL: openssl OpenSSL 1.1.1d-freebsd  10 Sep 2019` during post install. I have tried to install this plugin a several times and sometimes it works, my guess is that there is something wrong with the entropy as this seems to hang up during some DH param generation.~~ **Update**: Finally found the cause of this plugin installation getting stuck for a long time. The `openvpn` plugin enables the `ipfw` which locks out the job progress inside cirrus ci :smile:. A simple stop of this service after `post_install` made it pass.
  * ~~`quasselcore` UI not reachable~~
  * `unificontroller-lts` UI not reachable. Seems to be issues while starting up the controller after a clean installation. Created a new bug report for it here: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=254330

Related issues: 
* https://github.com/ix-plugin-hub/iocage-plugin-index/issues/180 (some more plugin 12.2 upgrades which would be needed to be synced to 12.2 branch later)
* Should fix the: https://github.com/ix-plugin-hub/iocage-plugin-index/issues/165 issue (there is some discussions about upgrading the mono version but this should solve the main installation problem) 
* `rtorrent-flood` specific issue: https://github.com/ix-plugin-hub/iocage-plugin-index/issues/138
* ~Old `gitlab` issue: https://github.com/ix-plugin-hub/iocage-plugin-index/issues/118. Starting to wonder if it's not a stale plugin.~ **Update**: Possible fix in: https://github.com/ix-plugin-hub/iocage-plugin-index/pull/185


fulder:
The currently failing tests on this branch are all known issues and the majority of them have PRs with fixes in plugin repositories. 

That said, the installation tests should be more stable and without UI false positives right now and I don't think I will do much more on this branch.

Hope we can merge: https://github.com/ix-plugin-hub/iocage-plugin-index/pull/178 (it is the one this branch/PR is based upon) now and I will remove the WIP status after this, meaning it should be ready for review/merge as well (although probably easier to do it after the https://github.com/ix-plugin-hub/iocage-plugin-index/pull/178 merge)

fulder:
@sonicaj Thank you for all the merges into master regarding my other PRs :heart:. 

Have merged upstream master into this one as well. New failures after merge/extra fixes:
* `deluge-pip`: Previously false positive, service was not started properly, PR/issue: https://github.com/ix-plugin-hub/iocage-plugin-index/pull/184
* `gitlab` Looks like there's a new (previously missed) bug with `gitlab_pages` not starting properly (will leave this for future plugin PRs to not block merging of this one any further)
* `rtorrent-flood` `rtorrent` service not supporting `status` due to missing `command` in service `rc.d` file. PR to fix this here: https://github.com/pierrickrouxel/iocage-plugin-rtorrent-flood/pull/9
* `zrepl` Service is not running, will also leave this one for future (plugin) PRs.

Have now removed the WIP status and this one should be ready for merge now as well :angel:

sonicaj (approved):
As always, thank you for your contribution :)

PR #181, opened 2021-03-04
Update 12.2 branch

sonicaj:

4. Add new plugin cirrus task


tprelog:
Sorry to be a nagging pain the ***  but plugins are failing to install with 12.1-RELEASE.

Please merge this so plugins will install correctly again

monica.json
```diff
@@ -6,31 +6,31 @@
...
+        "php74-dom-7.4.15",
+        "php74-gd-7.4.15",
+        "php74-gmp-7.4.15",
+        "php74-iconv-7.4.15",
+        "php74-intl-7.4.15",
+        "php74-json-7.4.15",
+        "php74-mbstring-7.4.15",
+        "php74-mysqli-7.4.15",
+        "php74-opcache-7.4.15",
+        "php74-pdo_mysql-7.4.15",
```
william-gr, line 25: This is likely to break very soon. Do we have to use specific version?

fulder: Good point, I added this as a quickfix for the already broken versions but also realized this will break after these packages are upgraded again. Can fix a new PR removing the versions.

fulder: Would this be better: https://github.com/ix-plugin-hub/iocage-plugin-index/pull/183 ?

william-gr: Yup, thanks!

william-gr (approved):

PR #184, opened 2021-03-07
Add dependency to deluge

aserrallerios:
I'm not very sure about this, but the daemon didn't start (could not find the libtorrent lib) until this dependency was installed.

Hopefully @jsegaert can shed some light.

### DIFF
--- a/deluge-pip.json
+++ b/deluge-pip.json
@@ -13,6 +13,7 @@
         "py37-pip",
         "py37-pillow",
         "py37-libtorrent-rasterbar",
+        "libtorrent-rasterbar",
         "rust"
     ],
     "packagesite": "http://pkg.FreeBSD.org/${ABI}/latest",


aserrallerios:
Actually, with this dependency the daemon starts, but the WebUI is unusable.

The 11-XX versions worked like a charm.

fulder:
> Actually, with this dependency the daemon starts, but the WebUI is unusable.

In what way is it unusable? (just wondering :)) I tried installing this plugin using your updated version (which seems like a good fix for the daemon being started properly again) but I could reach the UI and login using the default password. Or is there something else not working after that?

fulder:
> I'm not very sure about this, but the daemon didn't start 

As a side note, this is actually quite interesting why the new plugin installation tests passed for this plugin with the new version (see e.g.: https://cirrus-ci.com/task/6241764837163008) although installing it manually fails with this python dependency bug :thinking:

aserrallerios:
> In what way is it unusable? 

1. install the jail
2. console into the jail and pkt install libtorrent-rasterbar + service deluged start
3. Go to the WebUI
4. Add a torrent, any torrent

You'll notice that the download started (there's files in the Download folder of the jail), but the WebUI shows nothing, no torrents downloading. After a short period the UI will prompt: "The connection to the webserver has been lost!"

fulder:
Thank you for the explanation and the reproduce steps.

Seems like there are some more error logs regarding `libtorrent` after step 4):

```
...
Boost.Python.ArgumentError: Python argument types in
    torrent_handle.file_progress(torrent_handle)
did not match C++ signature:
    file_progress(libtorrent::torrent_handle {lvalue}, libtorrent::flags::bitfield_flag<unsigned char, libtorrent::file_progress_flags_tag, void> flags=0)
```

aserrallerios:
Good find. Probably we need an specific lib version, but a quick google search proved to be of no help.

fulder:
>As a side note, this is actually quite interesting why the new plugin installation tests passed for this plugin with the new version (see e.g.: https://cirrus-ci.com/task/6241764837163008) although installing it manually fails with this dependency bug :thinking:

FYI I tried to trigger the test again and got the same error now there as well: https://cirrus-ci.com/task/5795709669605376?command=install#L417 (still green but that's a known issue). This starts to look like there was some kind of dependency (package/pip package etc.) update which broken this plugin somehow.

I just found there were recent changes in `py-libtorrent-rasterbar` package some days ago, maybe that could be a hint/reason for what we are seeing: https://svnweb.freebsd.org/ports/head/net-p2p/py-libtorrent-rasterbar/?view=log

fulder:
Just tried to change the manifest packagesite from `latest` to `quarterly` and both the deamon and UI works better (could add a torrent file and it started to download in the UI as well, no errors in logs anymore either). 

Changed the packagesite in order to downgrade the `py37-libtorrent-restbar` package. See table here: https://www.freshports.org/net-p2p/py-libtorrent-rasterbar

jsegaert:
Hi Folks!

Some background.  I originally created the `deluge-pip` plugin to replace the regular `deluge` plugin.  The latter plugin just installed deluge-cli from pkg, but that package was broken for a very long time and hence `deluge-pip` was born as a workaround.
This may be a good opportunity to switch back to the regular deluge, i.e.  just install deluge-cli from pkg.  Unfortunately, it looks like that pkg suffers from the same dependency error, but when installing both `deluge-cli` and `libtorrent-rasterbar` it works fine.  I logged this as an [issue upstream](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=254118) so the package can be fixed.

Can one of you test with the following manifest file:
https://github.com/jsegaert/iocage-plugin-index/blob/fix_deluge/deluge.json

If you confirm it works, I can swap out `deluge-pip` back to vanilla `deluge`.

fulder:
Thanks for the explanation/story behind this plugin @jsegaert . I just tried to install it using your plugin but it still seems to fail with the previously mentioned C++ error: https://github.com/ix-plugin-hub/iocage-plugin-index/pull/184#issuecomment-792364915

jsegaert:
@fulder You are correct; the only "fix", is to install the older rasterbar-1.2.10 packages.

aserrallerios:
Thanks for taking care of this guys. If I'm not wrong, then it'd be enough to change fix the version of "py37-libtorrent-rasterbar"? (no need for "libtorrent-rasterbar")?

aserrallerios:
After wasting some time on this I was not able to downgrade `py37-libtorrent-rasterbar` to test it. I'm über freebsd noob, so I'd appreciate help on this (at least I'd like to have the jail manually fixed until it can be fully fixed).

Reading the discussion again it looks like:
* deluge-cli is outdate and suffers the same problems as pip
* we cannot install arbitrary versions of packages in jails

So we're screwed here? What can be done to at least workaround the problem for the community?

jsegaert:
As a not-very-pretty workaround, I backed up 1.2.10 versions of the `libtorrent-rasterbar` and `py37-libtorrent-rasterbar` packages to  a github repo so they don't get lost when the quarterly packages switch in a few weeks, and install them in the post_install script.

```
fetch https://raw.githubusercontent.com/jsegaert/iocage-plugin-index/fix_deluge/deluge.json
iocage fetch -P deluge.json -n deluge nat=1
```
While this hack seems to work, I'm not sure if this is pushing to limits too much.

aserrallerios:
Thanks, it works perfectly. If I understood correctly the pip version is newer? Should this workaround be applied to the pip version instead?

> While this hack seems to work, I'm not sure if this is pushing to limits too much.

IMO it's better to have a "hacky" plugin side that a broken plugin, WDYT @fulder ?

fulder:
Cool hack @jsegaert :shipit:, I agree @aserrallerios, although it's maybe not a very pretty hack, if it solves the current issues why not use it instead of having a broken plugin. 

We can always try to fix it in a more proper way if the issue @jsegaert created gets fixed or another workaround proposed.

jsegaert:
Cool, thanks for your votes of confidence.  I'll submit a PR in the next day or so.

Richardgriff (approved):

aserrallerios:
@Richardgriff probably it failed because the the "hack" was rollbacked https://github.com/jsegaert/iocage-plugin-deluge/commit/7e8ad30e4721a1908f8ce6491507f38b17751078

@jsegaert did you manage to fix it without the workaround?

jsegaert:
This PR is obsolete and has been superseded by #226 due to the move from Python 3.7 to 3.8.

aserrallerios:
So everything works good now? :)

fulder:
@aserrallerios I think it works much better with python 3.8. I just tried to manually install this plugin and follow the previous steps which previously made `deluged_web` crash. Looks like torrent files can be added and start downloading properly :)

The real fix for this issue was probably the revert of the `libtorrent-restbar` package to previous major version, see latest comment in the bug report created by @jsegaert : https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=254118#c2

aserrallerios:
Thanks for the detailed explanation!

PR #189, opened 2021-03-12
Update channels-dvr to use 12.2-RELEASE

tmm1:
Simple bump to 12.2-RELEASE

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -46,7 +46,7 @@ channels-dvr_task:
   only_if: "changesInclude('channels-dvr.json', '.cirrus/install_script.sh')"
   matrix:
     - freebsd_instance:
-        image_family: freebsd-11-3
+        image_family: freebsd-12-2
   env:
     PLUGIN_FILE: "channels-dvr.json"
 
--- a/channels-dvr.json
+++ b/channels-dvr.json
@@ -1,6 +1,6 @@
 {
     "name": "channels-dvr",
-    "release": "11.3-RELEASE",
+    "release": "12.2-RELEASE",
     "artifact": "https://github.com/fancybits/iocage-plugin-channels-dvr.git",
     "properties": {
         "nat": 1,
--- a/channels-dvr.json
+++ b/channels-dvr.json
@@ -12,7 +12,7 @@
         "security/sudo",
         "www/chromium"
     ],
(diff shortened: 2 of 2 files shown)


fulder:
We have some new automated tests for plugin installation in this repo now :) Could you bump the version inside the `.cirrus.yml` for this plugin as well?

tmm1:
Done

fulder:
> Done

Thank you! Looks like there is some errors with the installation check. I'm guessing the plugin packages need an update as well (https://cirrus-ci.com/task/5061771565203456?command=install#L280)

tmm1:
Which package repo is used?

In the past there were issues but it was fixed with a repo switch? https://github.com/ix-plugin-hub/iocage-plugin-index/pull/98#issuecomment-659801555

tmm1:
Looks like the json has it:

>    "packagesite": "http://pkg.FreeBSD.org/${ABI}/latest",

So I wonder why www/chromium is missing there?

tmm1:
@fulder the test is passing now!

Do I need to backport this to other branches so users on latest TrueNAS stable can use it?

fulder:
> Which package repo is used?
Looks like the json has it:
"packagesite": "http://pkg.FreeBSD.org/${ABI}/latest",

That's correct, so after the 12.2 bump in this PR this would resolve to: `http://pkg.FreeBSD.org/FreeBSD:12:amd64/latest`

>I need the chromium package and it's not available.
>So I wonder why www/chromium is missing there?

As you probably figured it out the chromium package seems to only be available in the `quarterly` packagesite (see table at: https://www.freshports.org/www/chromium). Not sure why it's missing for `FreeBSD:12:amd64` but is present for the 13/14 `latest` versions. Guess it is fine to use quarterly until it is released for the `latest` and/or we bump this to 13-14 again some time in the future. 

> @fulder the test is passing now!

Good job! Then I'm happy :smile: 

> Do I need to backport this to other branches so users on latest TrueNAS stable can use it?

In order for this change to take effect on the latest TN system it needs to be backported to the currently latest branch (`12.2-RELEASE`), yes.

fulder (approved):
Adding my approval :)

tmm1:
Thank you for the help! I opened a backport PR as well. Hope these can both be merged soon.

kmoore134 (approved):

PR #191, opened 2021-03-16
Upgrade unifi

fulder:
Looks like unifi6 has been released for other FreeBSD versions than only 14 some days ago: https://www.freshports.org/net-mgmt/unifi6/. Hopefully it should be fine to upgrade the package for this plugin now as well. 

Also upgraded the jail version from 11.3 to 12.2.


MalteHillmann:
Thank you for that. I'll give it a try soon.

kmoore134 (approved):

PR #192, opened 2021-03-16
Check service status in cirrus tests

fulder:
Add extra check to plugin installation tests validating if services started during `post_install` script execution are up and running (using `status`).

The implementation checks current services using `service -e` before and after post_install is run and then runs `status` on the `rc.d` scripts which weren't included in the initial `service -e` run.

The main reason for this extra check was the false positive test results where both the post_install script passed as well as the UI was reachable, although there was something wrong with the running service making it fail during a more complicated use case (for example using the UI more than just reaching it). Good examples of such false positive test is the `deluge-pip` plugin, test run: https://cirrus-ci.com/task/4527133297999872 (related PR/issue: https://github.com/ix-plugin-hub/iocage-plugin-index/pull/184).

### DIFF
--- a/.cirrus/install_script.sh
+++ b/.cirrus/install_script.sh
@@ -12,12 +12,12 @@ check_service_status()
 
   echo "${services_after}" | while IFS=' ' read -r a
   do
-    if echo "${services_before}" | grep "$a"
+    if echo "${services_before}" | grep "${a}"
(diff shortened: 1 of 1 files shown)


sonicaj (approved):
Thank you @fulder for your contribution :)

PR #197, opened 2021-03-17
Ignore services check support

fulder:
Add new env var: `SKIP_SERVICE_CHECK` for skipping service check for plugin tests. 

A plugin example where this is needed is the `zerpl` plugin containing a config file which should be edited by the user. Before that is done the service won't start correctly. For more info see the `post_install` text in the plugin repo: https://github.com/johnramsden/iocage-plugin-zrepl/blob/master/post_install.sh#L25

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -537,6 +537,7 @@ zrepl_task:
         image_family: freebsd-12-1
   env:
     PLUGIN_FILE: "zrepl.json"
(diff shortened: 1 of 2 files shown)


fulder:
`zrepl` run seems green after this change. Feel free to review/merge this @sonicaj

sonicaj (approved):

PR #201, opened 2021-03-18
Remove unificontroller-lts plugin

fulder:
Remove the `unificontroller-lts` plugin. Although the package is called `LTS` it seems not as supported/maintained as the `unifi6` package. See issue about this here: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=254330 (where `unifi6` seems to be the recommended/maintained one). The LTS is currently running an old 5 version of the controller which is not starting up correctly so this plugin is broken right now anyway.

---

4. Add new plugin cirrus task


sonicaj (approved):
So it seems that the port has not been updated but it's still supported upstream by UniFi. Anyways, until the port has been updated, we can have this removed. Thank you for looking into it.

PR #202, opened 2021-03-18
Upgrade 12.1 plugins to 12.2

fulder:
* Bump all plugins with 12.1 versions (which is EoL) to 12.2
* Format all plugin manifests to JSON with 4 space indent


sonicaj (approved):
Thanks!

PR #203, opened 2021-03-20
Swap rslsync icon + rename ui to match old managed plugin

brianannis:
- Update icon to get rid of weird white border
- Rename plugin to match previous naming scheme for old btsync / resilio plugin


fulder (approved):
Looks like the name was already updated on `master` so this is just an icon change from what I can see? LGTM, can add an approval.

kmoore134 (approved):

PR #204, opened 2021-03-22
Add NxFilter plugin

xTITUSMAXIMUSX:
This is my first plugin. Let me know if I need to do anything else.


nxfilter.json
```diff
@@ -0,0 +1,23 @@
...
+  "pkgs": [
+        "curl",
+        "openjdk8-jre"
+        ],
+  "packagesite": "http://pkg.FreeBSD.org/${ABI}/quarterly",
+  "fingerprints": {
+                  "plugin-default": [
+                                  {
+                                  "function": "sha256",
+                                  "fingerprint": "b0170035af3acc5f3f3ae1859d>
```
fulder, line 18:
```suggestion
                                  "fingerprint": "b0170035af3acc5f3f3ae1859dc717101b4e6c1d0a794ad554928ca0cbb2f438"
```

nxfilter.json
```diff
@@ -0,0 +1,23 @@
...
+                          }
+                        ]
+        },
+        "official": "Community"
```
fulder, line 22: Think this should be `"official": false`

nxfilter.json
```diff
@@ -0,0 +1,23 @@
...
+}
```
fulder, line 23: Could you format this JSON with 4 spaces? :) Seems a bit odd right now.

nxfilter.json
```diff
@@ -0,0 +1,23 @@
...
+  "artifact": "https://github.com/xTITUSMAXIMUSX/iocage-plugin-nxfilter.git",
+  "properties": {
+        "allow_raw_sockets":"1",
+        "boot":"on"
+          },
+  "pkgs": [
+        "curl",
+        "openjdk8-jre"
+        ],
+  "packagesite": "http://pkg.FreeBSD.org/${ABI}/quarterly",
```
fulder, line 13: Are you using `quarterly` for some special reason, or isn't `latest` better in most cases?

fulder (changes requested):
Thank you for contributing and good job with a first plugin. I have a few small change requests.

The plugin installation tests are currently failing but seem like it's because of the JSON syntax error (see one of the review comments). I managed to install your plugin manually without any issues so should probably pass and get green after that fix.

nxfilter.json
xTITUSMAXIMUSX: Yes. Latest would be best. I have that updated. Thanks for pointing that out.

nxfilter.json
xTITUSMAXIMUSX: Yeah it was my first attempt. I think I got it formatted with 4 spaces now.

nxfilter.json
xTITUSMAXIMUSX: Ok I updated that section to false

fulder: I can't see this update yet? :worried:

fulder (approved):
Looks good to me now and the tests are green. Adding my approval :)

sonicaj (approved):
Thank you for your contribution!

PR #210, opened 2021-03-29
UrBackup Plugin

Tostis:
UrBackup Plugin based on official [instructions](https://www.urbackup.org/freenasserverinstall.html).

Artifact repo: https://github.com/Tostis/iocage-plugin-urbackup


urbackup.json
```diff
@@ -0,0 +1,24 @@
...
+    "fingerprints": {
+        "plugin-default": [
+            {
+                "function": "sha256",
+                "fingerprint": "b0170035af3acc5f3f3ae1859dc717101b4e6c1d0a794ad554928ca0cbb2f438"
+            }
+        ]
+    },
+    "official": false,
+    "revision": "0"
```
sonicaj, line 23: Can we please use an integer value here ( so `0` without the quotes ) ?

urbackup.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "UrBackup",
+    "plugin_schema": "2",
```
sonicaj, line 3: Same here

Tostis:
I've fixed the numbers. I'm sorry for wrong commit date.

INDEX
```diff
@@ -482,6 +482,16 @@
 		"official": false,
 		"primary_pkg": "unifi5"
 	},
+	"urbackup": {
+		"MANIFEST": "urbackup.json",
+		"name": "UrBackup",
+		"icon": "urbackup.png",
```
sonicaj, line 488: Can you please change this to `https://icons.freenas.org/community-icons/urbackup.png` as it's supposed to be a URI

sonicaj (changes requested):

urbackup.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "UrBackup",
+    "plugin_schema": 2,
```
fulder, line 3: The `plugin_schema` has been removed from the plugin schema in https://github.com/ix-plugin-hub/iocage-plugin-index/pull/220 so guess we can remove it here now as well (and good to merge the latest `master` into this to see if the manifest looks good according to the latest `iocage` schema.

INDEX
Tostis: I've changed the url. How does it work? Who puts the image on that url?

sonicaj (approved):
Thank you @Tostis for your contribution

PR #215, opened 2021-04-04
Update emby.json with https port in NAT

Baenwort:
adding the https port per: https://support.emby.media/support/solutions/articles/44001159258-advanced-menu to the NAT_Forwards. Changing the name to reflect the difference between this and the other port.

I would like to hear comments about making this the beta channel plug-in to make it usefully different from the other emby plug-in?

### DIFF
--- a/emby.json
+++ b/emby.json
@@ -5,7 +5,7 @@
     "official": false,
     "properties": {
         "nat": 1,
-        "nat_forwards": "tcp(8096:8096)","tcp(8920:8920)"
+        "nat_forwards": "tcp(8096:8096),tcp(8920:8920)"
     },
     "pkgs": [
         "emby-server"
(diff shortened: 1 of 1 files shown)


emby.json
```diff
@@ -1,11 +1,11 @@
 {
-    "name": "Emby",
+    "name": "Emby NAT Server",
     "release": "12.2-RELEASE",
     "artifact": "https://github.com/freenas/iocage-plugin-emby.git",
     "official": false,
     "properties": {
         "nat": 1,
-        "nat_forwards": "tcp(8096:8096)"
+        "nat_forwards": "tcp(8096:8096)","tcp(8920:8920)"
```
fulder, line 8: Looks like a small typo here. Should probably be `"nat_forwards": "tcp(8096:8096),tcp(8920:8920)"` instead.

PR #216, opened 2021-04-06
Update emby.json with ffmpeg fix

Baenwort:
Re-adding ffmpeg to fix transcoding errors until the artifact.git can be redirected to use emby_servers custom FFMPEG build. https://github.com/freenas/iocage-plugin-emby/pull/5 is pending and once it is pushed ffmpeg can be removed from the plugin json.

This will repair the following issues: https://www.truenas.com/community/threads/emby-new-movies-not-playable-after-upgrade-to-truenas-12.92303/ and https://www.truenas.com/community/threads/emby-plugin-no-playback-error.71961/ and and others on the emby forum.

### DIFF
--- a/emby.json
+++ b/emby.json
@@ -8,7 +8,8 @@
         "nat_forwards": "tcp(8096:8096)"
     },
     "pkgs": [
-        "emby-server"
+        "emby-server",
+        "ffmpeg"
     ],
     "packagesite": "http://pkg.FreeBSD.org/${ABI}/latest",
     "fingerprints": {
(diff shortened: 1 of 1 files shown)


emby.json
```diff
@@ -19,5 +20,5 @@
             }
         ]
     },
-    "revision": "1"
+    "revision": "2"
```
sonicaj, line 23: Let's please change this to an integer like `2` without the quotes

sonicaj (commented):
Btw is this required apart from the revision change ? ( this should cover the package change and also NAT port https://github.com/ix-plugin-hub/iocage-plugin-index/pull/218/files )

PR #220, opened 2021-04-22
Validate plugin schema

fulder:
Adds plugin manifest validation using a new schema in `iocage` repository. Changes:

* Update INDEX schema and use python `jsonschema` package for schema validation
* Use the new plugin manifest JSON schema from `iocage` repo (see PR: https://github.com/iocage/iocage/pull/1255) to validate all plugin JSON files
* Fix all (minor) issues with both INDEX and all plugin manifests

-----

4. Add new plugin cirrus task


INDEX
```diff
@@ -245,15 +246,15 @@
 		"icon": "https://icons.freenas.org/community-icons/madsonic.png",
 		"name": "MadSonic",
 		"official": false,
-		"primary_pkg": null
+		"primary_pkg": ""
```
sonicaj, line 249: Why are we changing this to empty string ?

sonicaj (commented):
Thank you @fulder as always for your contribution ;) Just one minor nit

INDEX
fulder: If I remember my thinking here correctly during this PR creation I was just looking at the `index.schema` (without really much changes to the properties) and solving the issue after that. Having the `primary_pkg` property set to `null` here will give the `jsonchema` exception (note that before this change the current INDEX check didn't capture these issues):

```
None: None is not of type 'string'
```

Another way to solve this could be to simply allow null values for this property and change it to `["string", "null"]` if you would prefer that?

sonicaj: Yes @fulder let's please do that as we would be changing behavior otherwise ( how it's working right now )

fulder: Sure thing, I've changed the schema to accept `null` values for `primary_pkg`, synced the branch from master and solved some new INDEX/plugin issues :) 

Looking through the `INDEX` file again I realize there are entries with both `null` values and empty strings, but you're right, better to keep the same behavior instead of introducing possible bugs to the index.

zrepl.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "zrepl",
-    "plugin_schema": 2,
```
sonicaj, line 3: @fulder let's remove this entirely, it's not being used anymore

sonicaj (commented):
@fulder let's please have `revision` as a numeric identifier instead of being string based ( which it really is )

zrepl.json
fulder: That's true, can't see any references att all in `iocage` repo. Have removed it now.

fulder:
> @fulder let's please have revision as a numeric identifier instead of being string based ( which it really is )

@sonicaj The current plugin schema is fetched from the `iocage` repo and master branch, but I created a new branch on my fork changing the plugin manifest revision from string to number and pointed the github validation action to the raw manifest URL: https://raw.githubusercontent.com/fulder/iocage/plugin-schema-fixes/iocage_lib/plugin_manifest.json. 

Created a new PR for the update in `iocage` repo here: https://github.com/iocage/iocage/pull/1260

.github/workflows/validateindex.yml
```diff
@@ -6,9 +6,35 @@ jobs:
...
+      run: |
+        python -m pip install --upgrade pip
+        pip install jsonschema
+
+    - name: Validate INDEX JSON
+      run: jsonschema -i ./INDEX ./index.schema
+
+    - name: Validate plugins JSONs
+      run: |
+        wget https://raw.githubusercontent.com/fulder/iocage/plugin-schema-fixes/iocage_lib/plugin_manifest.json -O plugin_manifest.schema
```
sonicaj, line 26: Let's please change this to upstream iocage again

fulder: Thanks for the really fast merge in `iocage` repository :+1:. Changed back to upstream master again.

sonicaj (approved):
Thank you @fulder as always :)

PR #227, opened 2021-05-30
Add JDownloader plugin

kennedyoliveira:

4. Add new plugin cirrus task


kennedyoliveira:
How to trigger the test again? I did some adjustment on the starting script to account for the test.

jsegaert:
To trigger the tests again, you can create and push an empty/dummy commit, something like this:  
`git commit --allow-empty -m "Empty commit (trigger CI tests again)"`

Your project repo seems to be missing a `ui.json` file.  Also, test your rc.d script to ensure that the command `service jdownloader status` returns a proper output.

kennedyoliveira:
> To trigger the tests again, you can create and push an empty/dummy commit, something like this:
> `git commit --allow-empty -m "Empty commit (trigger CI tests again)"`
> 
> Your project repo seems to be missing a `ui.json` file. Also, test your rc.d script to ensure that the command `service jdownloader status` returns a proper output.

Thanks for the info, I added the `ui.json` file and triggered the build again, although it's still fail.

Having trouble debugging it, locally it does work, but it takes a bit of time until it's up, the service starts and download the actual application and just after it restarts and then get the correct pid and return on `service jdownloader status`.

Is it possible to get some other infos from the CI? would like to check if it's downloading correctly in `/usr/local/jdownloader` and there should be some logging file there.

kennedyoliveira:
Ok, I noticed it downloads around 75 MB worth of updates and seems the CI internet is slow so it takes too much time and times out, updated the script to manually set the pid of the process while updating do now it works as expected .

fulder:
> Is it possible to get some other infos from the CI? would like to check if it's downloading correctly in /usr/local/jdownloader and there should be some logging file there.

This is actually something I've been thinking of as an improvement for the tests after the first implementation. Would be nice to allow custom "post common test scripts" to be run for a plugin where the maintainer would like to check things like you just described here and that's only specific for one plugin.

You're always welcome to include any changes or new features to the test script in a PR as well btw, it's inside this repo after all: https://github.com/ix-plugin-hub/iocage-plugin-index/blob/master/.cirrus/install_script.sh

fulder:
> Ok, I noticed it downloads around 75 MB worth of updates and seems the CI internet is slow so it takes too much time and times out, updated the script to manually set the pid of the process while updating do now it works as expected .

Nice job fixing the issue for the slow network, one could argue that if the CI network is slow the same issue could happen on a user TN machine, i.e. in this case making the plugin installation complete but the plugin not being up and working before the service is really up and running. In these cases, I think, it's probably better to fix such issues on the plugin level.

Another solution could be to wait for the service to be properly started (something like the CI tests are doing) inside the `post_install` instead of the `pid` fix.

kennedyoliveira:
> Another solution could be to wait for the service to be properly started

I thought about that too, but I was afraid to wait too long and leave the user without understanding, sure if the thing doesn't pop up faster the user might get confused as well about what is going on, but knowing at least that it installed successfully I think is better than waiting on the installation.

I installed some plugin that waited forever, don't remember which one, but I got confused, reloaded the page, tried again, and made a mess, so I went with the PID just so the user will know the software was installed and is booting up.

sonicaj (approved):

sonicaj:
Thank you @kennedyoliveira for your contribution

kennedyoliveira:
When does it show up in truenas? I have refreshed the community list but it doesn't show.

fulder:
@kennedyoliveira The master branch need to be ported back to the `12.2-RELEASE` branch in this repo for it to show up at the community list automatically.

kennedyoliveira:
Ah that makes sense. I have a few more plugins I want to contribute, should I target the master or the proper release I used?

Also, when does the master get merged back to the releases? Is it periodically?

fulder:
@kennedyoliveira It's better to target `master` and wait for the port back to release branch done by the repo maintainers IMO. It's done from time to time and when `master` starts to get many commits over the latest release branch. There was just a new PR created for such a port back to the `12.2-RELEASE`: https://github.com/ix-plugin-hub/iocage-plugin-index/pull/235 after this gets merged you should be able to refresh your community index and get all the new plugins (including this one).

Issue #230, opened 2021-05-30
 Fresh Plugin of Bazarr fails on install


pavak1:
error given on creation
"  bazarr had a failure Exception: RuntimeError Message: pkg error: - py37-pip :,py37-libxml2 : Refusing to fetch artifact and run post_install.sh! Partial plugin destroyed"

it appears that the plugin try's to install py37 vs py38 and py37 is unavailable

fulder:
@pavak1 thanks for the issue, good finding and timing. I was just looking through some of the plugins with old python 3.7 dependencies yesterday and created a PR for all missing plugins (without any open PR already): https://github.com/ix-plugin-hub/iocage-plugin-index/pull/232

I've also updated the bazarr plugin with more generic `python3` references, it has been merged and ported back to the 12.2 release branch: 
* https://github.com/fulder/iocage-plugin-bazarr/pull/6
* https://github.com/fulder/iocage-plugin-bazarr/pull/7

After these fixes the cirrus tests seemed to pass on my fork (see: https://cirrus-ci.com/task/6672071733280768) so should probably pass in this repo PR as well soon.

pavak1:
@fulder Thank you for looking into the problem and doing the work! It appears that the changes where blocked for merging issue #232, but probably just needs a review.

fulder:
@pavak1, I've pushed a small fix after a comment to the PR, but otherwise it's just waiting for approval and merge now :) 

As a side note it would be nice to capture these package issues faster and in a more automated manner (see: https://github.com/ix-plugin-hub/iocage-plugin-index/issues/194)

achmetinternet:
I installed the plugin and can SSH into the jail but the UI doesn't work, I get 

Internal Server Error
The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.

fulder:
@achmetinternet (this should maybe be another issue than this close one but answering here anyways) it looks like bazarr plugin has changed a few things how it is installed and updated. The Internal Server Errors you are seeing are probably because there are some assets files missing while just cloning the repo. I've merged and backproted some fixes (and quickfixes) which should make a current bazarr installation upgrade and a clean update of the plugin work better (I hope).

Could you test once again on your side if it works better or if there are some more issues with this plugin? Thank you for your help and information about this.

PR #232, opened 2021-05-31
Upgrade python to 3.8

fulder:
Upgrade plugins (with no opened PRs) from old python 3.7 version to 3.8 and add the `python3` packages for easier plugin maintenance (where python3 binaries can be used instead of upgrading them for every minor python release)

Closes: https://github.com/ix-plugin-hub/iocage-plugin-index/issues/230

-------

4. Add new plugin cirrus task


homeassistant.json
```diff
@@ -30,6 +30,7 @@
         "libxml2",
         "libxslt",
         "pkgconf",
+        "python3",
```
tprelog, line 33: Home Assistant has specific requirements for the Python version that is used.

- https://github.com/home-assistant/architecture/blob/master/adr/0002-minimum-supported-python-version.md
- https://github.com/home-assistant/architecture/blob/master/adr/0016-home-assistant-core.md

Also, the virtualenv needs to be reinstalled when Python is updated. I will need to refactor some of the plugin for this change to be automagically handled correctly. I've considered rebuilding the virtualenv every time a user runs the plugin update, however some users have reported this can take almost an hour. It really depends on the NAS hardware and number of integrations a user has added to their Home Assistant.

fulder: Thanks for the detailed explanation @tprelog. I added the `python3` package here (and in other plugin manifests) in order to make it easier for the maintainers to be able to support minor updates without need to changing anything at the plugin level. But realize it can be problematic when it comes to the 2 supported python versions you're describing and the homeassistant plugin (currently 3.9 and 3.8 if I understand it correctly: https://github.com/home-assistant/core/blob/d829df332d8353661268f45300ac919dd8eb5116/homeassistant/const.py#L11 ?).

During the bump I was checking the `post_install` scripts for the plugins to see if they are using the `python3` binary or e.g. already supporting the current `python3.8`. That said, this change should probably work right now (as the `python3` package is currently `python3.8`), but the issue is if it gets updated too fast when homeassistant core is not supporting the new one yet, do I understand this correctly? If so, this would mean this package shouldn't be used by the plugin post install anyways and shouldn't be needed for this plugin (in this case I can remove it again).

fulder: I've removed the package again. Better to not have it there if it can cause more problems than it would "solve".

tprelog: So yea, Home Assistant currently supports python 3.8 and 3.9. When the next minor version of Python is released support for 3.8 will be dropped. However it can take a few  months before the new version is fully supported.

There are other minor nuance as well. One example, when the plugin was using python 3.7, there was a package `py37-av` but `py38-av` is not available.

In the perfect world, we would see a method for implementing a plugin branch (or version) as we previously discussed several months ago (I just now realized I somehow missed the notifications [when you had tagged me](https://github.com/iocage/iocage/pull/1248), sorry about that)

I think Home Assistant is an edge case that really pushes the boundaries of what is currently possible within the iocage plugin system and a perfect example of why most experienced users would still recommend a standard jail over plugins but that discussion is off topic here.

sonicaj (commented):
@fulder can you please confirm if we are not regressing for `esphome` ?

homeassistant.json
fulder: Hard and locked dependencies like that are never fun to work with :/ Good job maintaining this plugin even though it is this complex :+1: 

No problem about the versioning PRs/discussions in the `iocage` repo, reviews and comments are always much appreciated when you get some free time, but no stress at all :).

I will resolve this discussion as this PR won't change anything with Home Assistant plugin, feel free to open it up again if there's something more to discuss here.

fulder:
@sonicaj, from my understanding `esphome` should be ready for this change. The current cirrus task is failing due to the plugin being hardcoded to the `python3.7` version. Once this is merge, the plugin PR here should be merged as well: https://github.com/tprelog/iocage-esphome/pull/7 or is there something I've missed regarding this plugin (like Home Assistant) @tprelog (I'm guessing it is fine because of your comment in the PR, just doublechecking :smiley_cat: ) ?

sonicaj (approved):
Thank you for clarifying @fulder, let's wait for @tprelog to confirm/merge and we can merge this asap as well then :)

tprelog:
@sonicaj - Let's go ahead and merge this.

I can merge @fulders pull request in the ESPHome repo as well

sonicaj:
Awesome, thank you @tprelog for confirming

tprelog:
I just merged @fulder pull request on my end for ESPHome

fulder:
Thanks for the merges! :)

Issue #194, opened 2021-03-17
Run cirrus tests periodically


fulder:
I've been trying to make the new cirrus tests more stable and remove any false positives I could think of :crossed_fingers:. But was thinking it could be a good idea to run the cirrus tests (from master) e.g. every night in order to catch bugs such as broken packages or dependency issues before any users installing the plugins will.

What do you think about this idea? 

Created an issue instead of PR as any cirrus cron jobs need to be enabled in the cirrus ui (https://cirrus-ci.com/) inside project settings and cannot be done using the `.cirrus.yml`. See: https://cirrus-ci.org/guide/writing-tasks/#cron-builds

tprelog:
I agree, this would be nice to have. If not nightly, then at least on weekly basis

jsegaert:
I also agree; almost always breakage to the plugins I maintain is caused by some upstream changes in some packages, and not by a PR or change to the plugin itself.

Are those (very cool) CI tests still active?  Somehow my latest PR #222 didn't seem to trigger any tests...

fulder:
@jsegaert the tests should still be active. I'm confused why it didn't run on your PR. Just tried to both create a test MR on my fork with a small change in the same `mineos.json` and it triggered just fine (https://cirrus-ci.com/build/4731071641157632) as well as sync upstream master to one of my currently open PRs here which also triggered the correct task: https://github.com/ix-plugin-hub/iocage-plugin-index/pull/208/checks?check_run_id=2533775947

If I understand the cirrus `changesInclude` function correctly, it should trigger the `mineos` task if the PR has introduced some new changes to the `mineos.json` file (see: https://cirrus-ci.org/guide/writing-tasks/#supported-functions)

Maybe there was some sort of hickup between the CI and GitHub, could you maybe try pushing some dummy commit to your branch and see if it still doesn't work?

fulder:
@kmoore134 and @sonicaj, any thoughts on enabling scheduled tests for this repository in order to catch dependency issues faster? :) (I have a dream (:smile:)  the tests are close to getting all green and once a new dependency is no longer working it would toggle to red status on the `master` branch)

jsegaert:
Also, @kmoore134  and @sonicaj, there are a lot of changes/fixes in `master`, is there any timeline on merging these into `12.2-RELEASE` so the TrueNAS can benefit from our efforts?  😀

sonicaj:
[...]

PR #242, opened 2021-06-21
update pkgs for zwavejs2mqtt

tprelog:
Update the packages for Z-Wave JS to MQTT
 - adds `jq` and `yarn`
 - switch to current `node` version

---

<s>On a side note @fulder - I looked back at the cirrus results and the builds have been passing even though the service failed to start and the admin portal was not reachable. Should the cirrus test have caught this? If so, I think it should have been failing for a few weeks now -- Or will it pass because the jail itself is successful and running?

https://github.com/tprelog/iocage-zwavejs2mqtt/issues/1</s>

Sorry @fulder, it looks like we are only running cirrus on a plug request (and merge)? I had myself confused with this issue #194 - I thought we were already running the periodic cirrus test, but now I realize this is not correct.

---

4. Add new plugin cirrus task

### DIFF
--- a/zwavejs2mqtt.json
+++ b/zwavejs2mqtt.json
@@ -22,9 +22,11 @@
     "pkgs": [
         "bash",
         "curl",
-        "node14",
-        "npm-node14",
-        "python"
+        "jq",
+        "node",
+        "npm",
+        "python",
+        "yarn"
     ],
     "packagesite": "http://pkg.FreeBSD.org/${ABI}/latest",
     "fingerprints": {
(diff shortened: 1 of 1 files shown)


fulder:
@tprelog you're correct, the tests are not running periodically right now (would be nice if they did), so hard to catch these issue in a more automatic way. Currently tests for a specific plugin will only be run if either the plugin manifest or the cirrus tests file itself is changed. 

If this plugins wasn't recently updated there are no current runs for it either :/ Which is bad and one of the reason for creating issue: #194

PR #244, opened 2021-06-26
Add redis plugin

kennedyoliveira:

4. Add new plugin cirrus task


fulder:
LGTM and both checks are green :+1: just a minor (and stupid) question: Whats the reason for installing redis from source rather than using the already prebuilt `redis-6.0.14` port (except it being a bit older than the source version)?

kennedyoliveira:
The question is not stupid at all, but you already answered yet , it's for easy of update and keeping on the latest version.

redis.json
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "Redis",
+    "release": "12.2-RELEASE",
+    "artifact": "https://github.com/kennedyoliveira/iocage-redis-plugin",
+    "properties": {
+        "dhcp": 1,
+        "boot": "on"
```
sonicaj, line 7: This is not really required and is on by default

sonicaj (changes requested):
@kennedyoliveira you forgot to add redis to INDEX ;)

redis.json
kennedyoliveira: ah ok, removed.

kennedyoliveira:
> @kennedyoliveira you forgot to add redis to INDEX ;)

True lol, should not do this stuff when you are almost falling asleep, added, thanks!

sonicaj (approved):
Thank you for your contribution @kennedyoliveira!

Issue #236, opened 2021-06-10
Jackett Doesn't Have Its Port Forwards Set By Default 


jasker5183:
On a new default plugin jail the Jackett admin interface is unreachable because the port forwards aren't set to 9117. Easy enough to fix if you're aware of what's wrong.

fulder:
@jasker5183 thanks for this issue, I just created a PR for jackett nat properties in #237. Probably better to default to NAT settings with proper port forwarding in order to get it working for that kind of installation (changing it to DHCP should still work if wanted).

PR #268, opened 2021-09-20
Adding AdGuardHome Plugin

leandroscardua:
- Adding on CI
- AdguardHome Json
- adding on INDEX
- Adding Logo


adguardhome.json
```diff
@@ -0,0 +1,24 @@
...
+    "name": "adguardhome",
+    "release": "12.2-RELEASE",
+    "artifact": "https://github.com/leandroscardua/iocage-plugin-adguardhome.git",
+    "official": false,
+    "pkgs": [
+        "adguardhome"
+    ],
+    "properties": {
+        "dhcp": 1,
+        "boot": "on"
```
fulder, line 11: I think `boot` should be on by default so shouldn't be needed here. See same comment e.g. here: https://github.com/ix-plugin-hub/iocage-plugin-index/pull/244#discussion_r668820031

adguardhome.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "adguardhome",
+    "release": "12.2-RELEASE",
+    "artifact": "https://github.com/leandroscardua/iocage-plugin-adguardhome.git",
+    "official": false,
+    "pkgs": [
+        "adguardhome"
+    ],
+    "properties": {
+        "dhcp": 1,
```
fulder, line 10: Do you really need `dhcp` here or could it be changed to `NAT` instead? I do like DHCP for all my jails and use it for all of them but defaulting to `dhcp` without any `nat` forwards set can cause issues like: https://github.com/ix-plugin-hub/iocage-plugin-index/issues/236

adguardhome.json
leandroscardua: Ok, I`m going to remove it.

fulder (approved):
Your icon seems to have size 200x200 instead of the 128x128 specified in the PR steps (not sure if that still matters much, there are a lot of different sized icons, but can still be worth to make it smaller)

Other than that I added 2 minor comments, nothing big so adding my (optional) approval. 

Cirrus CI tests are (finally) green again so I'm happy :smile:  (Thanks again for your debug help)

adguardhome.json
leandroscardua: As the AdGuardHome works filter DNS query, using it as nat, I`m afraid to overloded the Truenas main IP and avoid conflict with other pluings like Bind DNS and Dnsmasq already on the ecosystem.
by default, the Adguardhome uses port 53, to receive the dns query, furthermore if the user changes that port to anything else and break the access to DNS.

yes, I can change it something like that? or not? or no need to the webpart?

        "nat": 1,
        "nat_forwards": "tcp(80:80)"    ## Web UI
        "nat_forwards": "tcp(53:53)"   ## DNS Query

fulder: That could "work" (just the `nat_forwards` property should be specified only once, if more mappings are needed they should be comma separated inside the value so something like: `"nat_forwards": "tcp(80:8585),tcp(53:53)"`. 

Although, like you said before this will collide with the `bind` plugin (if someone tries to install both of them), which can be confusing to debug when it happens.

Furthermore, setting a NAT forward for DNS would be problematic as usually (and to my knowledge) the OS DNS resolvers doesn't support other ports than the default `53`. That said I think changing this to `nat` could cause more potential issues than it really solves, so keep it with `dhcp` after all. Sorry for the confusion.

leandroscardua: Yes, I agree with you on that matter.
Let`s to it with dhcp

I hope everything is good now and we can apply it to the master branch :D

leandroscardua:
HI @fulder ,

I`m going to cancel that pull request, I made a mistake during the merge request.

PR #280, opened 2021-11-06
Update for homeassistant and zwavejs2mqtt

tprelog:
**Home Assistant**

Add packages for Python 3.9 support to resolve the following issue
- https://github.com/tprelog/iocage-homeassistant/issues/47

**ZWaveJS 2 MQTT**

Switch back to Node 14 to resolve the following issue
- https://github.com/tprelog/iocage-zwavejs2mqtt/issues/3

Node 16 would be better, but `npm-node16` is not available in the package repo


sonicaj (approved):
Thanks!

PR #292, opened 2022-02-15
New iocage plugin for Prowlarr

jasker5183:
Created iocage plugin for Prowlarr,

1. Added Prowlarr entry to INDEX in alphabetical order.
2. Added a new icon in the ./icon/ directory, prowlarr.png.
3. Ensured the submitted icon file is a valid PNG that is 128x128 in size.
4. Added new plugin cirrus task, prowlarr_task.

Everything should be all there. Might want to clone the artifact repo some where else though.


prowlarr.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "Prowlarr",
+    "release": "12.2-RELEASE",
```
sonicaj, line 3: Let's please use `12.3-RELEASE` instead

.cirrus.yml
```diff
@@ -465,6 +465,15 @@ privatebin_task:
   env:
     PLUGIN_FILE: "privatebin.json"
 
+prowlarr_task:
+  <<: *INSTALL_PLUGIN
+  only_if: "changesInclude('prowlarr.json', '.cirrus/install_script.sh')"
+  matrix:
+    - freebsd_instance:
+        image_family: freebsd-12-2
```
sonicaj, line 473: Same here

sonicaj (commented):
Thank you @jasker5183 for your contribution, a minor nit

jasker5183:
Switched to 12.3-RELEASE in requested files and now it fails. Unsure at the point of doing this as the base TrueNAS OS is still 12.2 and all the other plugins are also still 12.2.

sonicaj:
@jasker5183 do we know why it fails ? For a new plugin, i still think it would be nice if we could use 12.3 as 12.2 would be EOL soon.

jasker5183:
It's failing the cirrus test as you can see, something about `Service /etc/rc.d/cleartmp not started`. No idea what's going on there. 

As for the jail release version would it not just be simpler to leave it on 12.2 as TrueNAS is scheduled to update to 13.0 around April 12 (12.2 supported until March 31). After that point it would be better to update the jails to version 13.0-RELEASE.

Correct me if I'm wrong but if the base OS is on 12.2 and you're running a 12.3 jail you're still running the 12.2 kernel? Also not too sure how that works.

sonicaj:
@jasker5183 have you tested 12.3 working locally ? I am not sure if it's a CI issue or a plugin issue but as a maintainer, i think each plugin should be tested locally.
About the kernel bit, yes that's true but the vulnerabilities found in userland, you are safe/covered from that.

jasker5183:
Just tested it and I guess this is normal?

    Using RELEASE: 12.3-RELEASE
    Using Branch: 12.2-RELEASE

But yes it's working:

`29  | Prowlarr     | up    | 12.3-RELEASE | 172.16.0.22`

I guess the problem is with the CI.

sonicaj (approved):
Thanks!

callaa3:
Showing at https://www.truenas.com/apps/ but not in the comunity list of plugins on my truenas install despite hitting Refresh Index. Not quite ready? Or maybe I need to swtich train?

Only did the Truenas clean install a month ago and the config seems to be re-loading OK...

Mar  7 00:00:00 truenas syslog-ng[1015]: Configuration reload request received, reloading configuration;
Mar  7 00:00:00 truenas syslog-ng[1015]: Configuration reload finished;

Current Train: TrueNAS-12.0-STABLE - Release Train for TrueNAS 12.0 [release]

jasker5183:
Please see #299 and #295, basically the 12.2 branch needs to be updated in order for the new plugin's to show up in the TrueNAS GUI.

PR #296, opened 2022-02-16
Fix INDEX Formatting

jasker5183:
Fix NexusOSS formatting
Remove extra line

There is something wrong with the INDEX file as far as I can tell starting with line 381, these edits "seem" to fix it. When editing there is also an extra line at the bottom, I tried deleting it but I think it's still there. Should line 659 be `},`? 

If this is a waste of time just reject and close, I was just looking for a reason why my TrueNAS box hasn't pulled down the last four plugin updates and thought this might have something to do with it.

### DIFF
--- a/INDEX
+++ b/INDEX
@@ -378,15 +378,15 @@
 		"name": "Netdata",
 		"official": false
 	},
-    "nexus-oss": {
-        "MANIFEST": "nexus-oss.json",
-        "name": "NexusOSS",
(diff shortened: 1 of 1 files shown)


sonicaj (approved):
Thank you @jasker5183 for your contribution!

sonicaj:
@jasker5183 this fixes the formatting indeed but this should not affect your box's ability to pull down latest changes.

PR #304, opened 2022-03-09
Updated unificontroller to version 7

sdalmeida:
Bumped unificontroller to version 7


sdalmeida:
Even though unifi7 is available in freshports, it seems like it hasn't been updated in http://pkg.FreeBSD.org

PR #310, opened 2022-04-26
Archiware P5 plugin for TrueNAS

IgorC-test:
Hello, I would like to submit new plugin for TrueNAS, Archiware P5 (https://www.archiware.com/release-notes), it has tape support so it needs devfs rules to pass appropriate devices.
I haven't uploaded icon since I do not have the upload rights and didn't modify INDEX before the initial submission
Thank you for your support

4. Add new plugin cirrus task


archiwarep5.json
```diff
@@ -0,0 +1,46 @@
...
+        "mount_fdescfs": "1",
+        "allow_mount": "1",
+        "allow_mount_devfs": "1",
+        "allow_mount_procfs": "1",        
+        "allow_raw_sockets": "1",
+        "dhcp": "1",
+        "securelevel": "0",
+        "enforce_statfs": "0"
+    },
+    "release": "12.2-RELEASE"
```
sonicaj, line 45: Let's please use `12.3-RELEASE` as 12.2 is already EOL

sonicaj (commented):
Thank you @IgorC-test for your contribution. Could we update INDEX next now ?
Also curious if we can find some image which is okay to be uploaded here ?

sonicaj (changes requested):
Can we please fix the JSON ( see the CI test please ) ?
Also let's please test this before having us merged this to ensure it actually works. Thanks!

sonicaj (commented):
@IgorC-test can you confirm please if you have verified the latest changes to ensure they work as desired ? Secondly i am not seeing the icon file - could you please upload it as well as the URI you have specified won't work otherwise. Thanks!

sonicaj (approved):
Thank you @IgorC-test for your contribution!

PR #327, opened 2022-08-18
Updating Emby Plugin

Baenwort:
I have updated the INDEX as part of turning the two Emby plug-ins into a useful pair of stable and beta/development branch plug-ins.

This is after asking the developers at: https://emby.media/community/index.php?/topic/13083-freenas-plugin/page/46/#comments and receiving their indication that it is a good idea.

This will also fix them for 13.1 and the switch from mono to .NET (adding mlock and updating the package list). I've also standardized on the VNET instead of NAT as that is how the original Emby plug-in works and I've discovered a bug that I'm in the process with a JIRA ticket 117272 to have corrected that using VNET works around.


sonicaj (commented):
@Baenwort can you please rebase your changes from master as currently we have a conflict ? Thanks!

Baenwort:
@sonicaj This should be ready now

sonicaj (approved):
Thank you @Baenwort for your contributions!

PR #331, opened 2022-09-17
remove unsupported plugins

tprelog:
[Home Assistant only supports Linux, OSX and Windows using WSL. 
](https://github.com/home-assistant/architecture/blob/master/adr/0016-home-assistant-core.md#0016-installation-method-home-assistant-core)

In light of recent changes to Home Assistant, it's time to have this plugin removed from TrueNAS CORE. As Home Assistant continues to grow and add new features, installation on FreeBSD no longer “just works” as it had for so many years in the past.

[esphome was removed back in June](https://github.com/ix-plugin-hub/iocage-plugin-index/pull/313) but the manifest snuck back in.

The `zwavejs2mqtt` project has been renamed to `zwave-js-ui` which broke the install script - It’s not a difficult fix but the plugin should also be renamed. Since I’m no longer using TrueNAS CORE or any iocage jails its not feasible for me to update and re-release this plugin under its new name.


sonicaj (approved):
Thank you @tprelog for your contributions!

PR #334, opened 2022-09-30
Add FlareSolverr plugin

pigfrown:
Adds FlareSolverr plugin.

Most of the work is done in the post_install.sh script because FlareSolverr isn't in the repository. This includes patching a dependency to work correctly on FreeBSD.


sonicaj (approved):
Thank you @pigfrown for your contribution!

PR #337, opened 2022-10-22
fixes #336: add collectd5 package required by InfluxDB

### LINKED ISSUE #336: InfluxDB unable to run
In the default configuration provided by the **grafana** plugin, InfluxDB is enabled, meaning that the InfluxDB server will monitor the files in the /usr/local/share/collectd directory.

This directory is present only if the `collectd5` package is installed.

I will provide a PR to fix this issue, introducing the dependency on collectd5.

Another option is to disable collectd in the default InfluxDB configuration.

alberanid:
Fix issue #336 adding the missing dependency (**collectd5**)

### DIFF
--- a/grafana.json
+++ b/grafana.json
@@ -5,6 +5,7 @@
     "pkgs": [
         "grafana8",
         "influxdb",
+        "collectd5",
         "curl"
     ],
     "properties": {


sonicaj (approved):

sonicaj:
Thank you @alberanid for your contribution!

PR #346, opened 2022-12-15
Fix Motioneye Plugin Python Usage

Baenwort:
This fixes the artefact for the motioneye plugin to remove fixed call of python 2.3 and made it use newest Python 2.7 but this is only start of moving to Python 3 now that motion package is moving forward again.


sonicaj (approved):
Thank you @Baenwort for your contribution!

PR #347, opened 2022-12-18
Python3 update of MotionEye

Baenwort:
This is the initial release of getting the MotionEye plug-in to use Python 3 (still in support) instead of Python 2.7 (past EOL) by moving to the dev branch of motioneye. 

This should be temporary until the .43 release of motioneye when I can switch back to the main branch.

I am also switching the artefact to a temp repo to facilitate so that there isn't any period of time after this PR is merged, but before I can update the main repo, that things won't work. Once this PR is accepted I will update the main artefact repo and then issue a PR to update the manifest.json to point back at it.

This is due to the 'branch' variable of iocage fetch not being supported in the manifest system. Maybe some day [this](https://github.com/iocage/iocage/pull/1248) PR will be accepted and the manifest system will acknowledge the branch variable that currently works with iocage due to the hard work of the FreeBSD port maintainer.

### DIFF
--- a/motioneye.json
+++ b/motioneye.json
@@ -10,6 +10,7 @@
         "nat_forwards": "tcp(8765:8765)"
     },
     "pkgs": [
+        "bash",
         "curl",
         "python3",
         "python39",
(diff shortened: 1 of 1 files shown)


sonicaj (approved):
Thank you @Baenwort for your contribution!

PR #351, opened 2022-12-29
Added rabbbitmq plugin

fadushin:
This PR adds a plugin for the RabbitMQ message broker.

RabbitMQ is installed and configured with an `admin` user with administrative permissions.

The management plugin is installed so that the broker can be configured via the web UI.

~~~

4. Add new plugin cirrus task


fadushin:
Please let me know if you need an icon with a transparent background.

rabbitmq.json
```diff
@@ -0,0 +1,22 @@
+{
+    "name": "rabbitmq",
+    "release": "13.1-RELEASE",
+    "artifact": "https://github.com/fadushin/iocage-plugin-rabbitmq.git",
+    "pkgs": [
+        "rabbitmq"
+    ],
+    "properties": {
+        "dhcp": 1
```
sonicaj, line 9: Does it has to be dhcp or can we default to `nat` and specify port forwards instead ?

fadushin: Interesting question.  RabbitMQ can speak many protocols, and it would be up to the user to subsequently enable plugins that use different ports (e.g., AMQP, MQTT, MQTT/HTTP, and TLS variants).  So to me it is a question of what the impact on the user will be if she or he decides to use more protocols later, or whether all of those protocols should be specified up front.

I don't have a clear sense of the impact, so I would go by your guidance.  Do you have any examples you can point me to that use NAT and port forwarding?

sonicaj: We would like to default to NAT and user can of course change this later. You can find an example here: https://github.com/ix-plugin-hub/iocage-plugin-index/blob/d1cf514cd6746bba6b8cb4f629ffa927aa40dd53/amule.json#L7

fadushin: Okay, let me do some tests.  I'll request a re-review when I push changes to this branch.

fadushin: Changed to NAT with forward porting of AMQP and HTTP interfaces.

fadushin: Also updated icon with bg transparency.

sonicaj (approved):
Thank you @fadushin for your contribution!

PR #355, opened 2023-01-27
Use elasticsearch7 in graylog plugin

erkmos:
The elasticsearch6 port has been deprecated and removed from the ports tree, this causes installation of this plugin to fail with 13.1. With this change we use elasticsearch7 instead

See: https://github.com/FreeBSD/freebsd-ports/commit/dc721d84ea9ad5e97d2c518b2bdca0fd37a4382b

### DIFF
--- a/graylog.json
+++ b/graylog.json
@@ -9,7 +9,7 @@
     "pkgs": [
         "graylog",
         "elasticsearch7",
-        "mongodb44",
+        "mongodb50",
         "pwgen",
         "p5-Digest-SHA"
     ],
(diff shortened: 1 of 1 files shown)


sonicaj (approved):
Thank you @erkmos for your contribution!

sonicaj (commented):
Actually please look at https://github.com/ix-plugin-hub/iocage-plugin-index/pull/355/checks?check_run_id=10933925930 - we have test failure

sonicaj (approved):
Thank you @erkmos

PR #356, opened 2023-01-30
Add PhotoPrism plugin

oridyn:

4. Add new plugin cirrus task


photoprism.json
```diff
@@ -0,0 +1,30 @@
+{
+    "name": "PhotoPrism-Test",
+    "release": "12.4-RELEASE",
```
sonicaj, line 3: Let's please use 13.1-RELEASE ?

photoprism.json
```diff
@@ -0,0 +1,30 @@
...
+    "artifact": "https://github.com/oridyn/iocage-plugin-photoprism",
+    "official": false,
+    "properties": {
+        "vnet": 1,
+        "dhcp": 1
```
sonicaj, line 8: We ideally want to default to nat, is there some sort of requirement by the application for dhcp ?

INDEX
```diff
@@ -401,6 +401,14 @@
 		"official": false,
 		"primary_pkg": null
 	},
+	"photoprism": {
+		"MANIFEST": "photoprism.json",
+		"description": "PhotoPrism® is an AI-Powered Photos App for the Decentralized Web. It makes use of the latest technologies to tag and find pictures automatically without getting in your way.",
+		"icon": "https://icons.freenas.org/community-icons/photoprism.png",
+		"name": "PhotoPrism®",
```
sonicaj, line 408: Why the discrepancy in the name defined in `photoprism.json` and here ?

sonicaj (commented):
Thank you @oridyn for your contribution!

A few minor nits

photoprism.json
oridyn: I'm afraid I can't.

At this point there is no pre-built binary for 13.1-RELEASE. Only for 12: https://github.com/psa/photoprism-freebsd-port/releases/tag/2022-11-18.

photoprism.json
oridyn: I've added `avahi-dameon` and `nginx` so that the plugin's UI can be accessed under `http://<jailname>.local` on port 80.
I prefer it to accessing through the IP and port.

Happy to change this though to NAT and port if you think that's better for these plugins.

INDEX
oridyn: Didn't realise there's a connection. Updated it now.

sonicaj (approved):
Thanks @oridyn!

PR #377, opened 2023-08-28
Update plugins to 13.2-release

spz2k9:
Hi,

Jellyfin plugin had a ffmpeg error and the Fix was to upgrade the release from 13.1 to 13.2. I thought I'd update all of 'em since doing one or 67 is the same time.

*Note : I've only replaced the release from 13.1 to 13.2 not any other versions and I didn't update the inner packages.

### DIFF
--- a/adguardhome.json
+++ b/adguardhome.json
@@ -1,6 +1,6 @@
 {
     "name": "adguardhome",
-    "release": "13.1-RELEASE",
+    "release": "13.2-RELEASE",
     "artifact": "https://github.com/leandroscardua/iocage-plugin-adguardhome.git",
     "official": false,
     "pkgs": [
--- a/backuppc.json
+++ b/backuppc.json
@@ -1,6 +1,6 @@
 {
     "name": "BackupPC",
-    "release": "13.1-RELEASE",
+    "release": "13.2-RELEASE",
     "official": false,
     "artifact": "https://github.com/freenas/iocage-plugin-backuppc.git",
     "properties": {
--- a/bazarr.json
+++ b/bazarr.json
@@ -1,6 +1,6 @@
 {
     "name": "bazarr",
-    "release": "13.2-RELEASE",
+    "release": "13.1-RELEASE",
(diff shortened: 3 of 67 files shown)


spz2k9:
@sonicaj Hi, I've seen you did a version bump last year. So I've added you so that the job doesn't get done twice. I've upgraded most packages (21 removed since they didn't build correctly).

EuanRiggans:
Hopefully this gets merged soon, the Jellyfin problem also effects Emby

jasker5183:
I don't think they can update the plugins to 13.2 until the underlying OS has been updated as well, which I hope is soon because 13.1 is EOL.

pmhausen:
TrueNAS does not run a specific release but 13-STABLE. 13.2-RELEASE jails are perfectly supported. Please update your plugins to a supported release. Breaking packages are expected if you don't.

jasker5183:
As of TrueNas 13-U5:

```

cat /etc/os-release 
NAME=FreeBSD
VERSION="13.1-RELEASE-p7"
VERSION_ID="13.1"
ID=freebsd
ANSI_COLOR="0;31"
PRETTY_NAME="FreeBSD 13.1-RELEASE-p7"
CPE_NAME="cpe:/o:freebsd:freebsd:13.1"
HOME_URL="https://FreeBSD.org/"
BUG_REPORT_URL="https://bugs.FreeBSD.org/"

```

They could update the jails to 13.2 but the jails would still be using the kernel from 13.1. I've also never seen them update the community plugins before the OS gets the update. I'm guessing that's why not one maintainer has responded to this PR yet.

pmhausen:
> They could update the jails to 13.2 but the jails would still be using the kernel from 13.1.

Yes. This is supported, because within a major -STABLE release no kernel ABI changes occur.

jasker5183:
Was looking for the maintainers of this repo (to maybe speed this up) and was shocked to find out that @sonicaj is the only one, hope it doesn't take too long before he gets to this PR.

janknieling (approved):

sonicaj (approved):

PR #379, opened 2023-09-15
Remove transmission-web

jasker5183:
Transmission has been updated to 4.0.4 and as such transmission-web has been included in the transmission-daemon package so is no longer needed and the update fails otherwise.

### DIFF
--- a/transmission.json
+++ b/transmission.json
@@ -8,8 +8,7 @@
         "nat_forwards": "tcp(9091:9091)"
     },
     "pkgs": [
-        "transmission-daemon",
-        "transmission-web"
+        "transmission-daemon"
     ],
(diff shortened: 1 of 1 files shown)


jasker5183:
Just tested the jail myself was able to reach the admin portal just fine, no idea why the self test failed unless it's not installing from scratch. Also for Transmission installs just upgrading they might fail because 3.00_5 doesn't conflict with 4.0.4 so you have to remove transmission-web manually.

perrin4869:
the package was removed [here](https://cgit.freebsd.org/ports/commit/?id=8da970ab75e1a1031d93ff8e1aa8e3ff71afef07) for more context

sonicaj (approved):
Thank you @jasker5183 for your contribution!

PR #400, opened 2024-10-14
Add Immich plugin

zebrapurring:
This change adds the possibility to install Immich v1.117.0 as an IOCage plugin via https://github.com/zebrapurring/iocage-plugin-immich.

The following command can be used to create an Immich jail while the PR is not integrated yet:

```sh
iocage fetch --git_repository https://github.com/zebrapurring/iocage-plugin-index --plugin-name immich --name immich_jail
```


kmoore134:
Just as a note, this repo is no longer monitored or maintained. These plugins for TrueNAS 13.0 are deprecated, and on 13.3 plugins (and jails) are expected to stop working in the near future. Users running Apps should schedule migration to TrueNAS SCALE to run the Docker versions.

